PR #20, opened 2018-11-06
Fix rootURI processing

reznikmm:
URI can be in two form: `file://absolute_path` and `file:relative_path`.
Process both form correctly. Fix rootURI in tests.


source/ada/lsp-ada_handlers.adb
```diff
@@ -65,10 +65,14 @@ package body LSP.Ada_Handlers is
       Response.result.capabilities.textDocumentSync :=
         (Is_Set => True, Is_Number => True, Value => LSP.Messages.Full);
 
-      if not LSP.Types.Is_Empty (Value.rootUri) then
```
raph-amiard: Some comments wouldn't hurt here, this bit of code is a bit hard to understand as it is to me

PR #22, opened 2018-11-06
First pass of support for completion

raph-amiard:
Depends on https://github.com/AdaCore/langkit/pull/102
and https://github.com/AdaCore/libadalang/pull/164

### DIFF
--- a/source/ada/lsp-ada_contexts.adb
+++ b/source/ada/lsp-ada_contexts.adb
@@ -16,11 +16,13 @@
 ------------------------------------------------------------------------------
 
 with Ada.Directories;
-with Ada.Strings.UTF_Encoding.Wide_Wide_Strings;
+with Ada.Strings.UTF_Encoding;
 with Ada.Text_IO;
 
 with GNATCOLL.JSON;
 with GNAT.OS_Lib;
+
+with Libadalang.Project_Provider;
 
 package body LSP.Ada_Contexts is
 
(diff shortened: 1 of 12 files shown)


reznikmm:
It seems we need to bump stable branch of libadalang to make it working

source/ada/lsp-ada_documents.adb
```diff
@@ -33,9 +33,10 @@ package body LSP.Ada_Documents is
    function To_LSP_String
      (Value : Wide_Wide_String) return LSP.Types.LSP_String;
 
-   function Get_Symbol_Kind
```
reznikmm: Why do you rename this function? It returns LSP.Messages.SymbolKind; So Get_Symbol_Kind is good enough for me.

source/ada/lsp-ada_contexts.adb
```diff
@@ -112,31 +124,44 @@ package body LSP.Ada_Contexts is
...
             Base_Name => GNATCOLL.VFS.Filesystem_String (Name));
 
          if Result.Is_Regular_File then
             return Result;
          end if;
-      end if;
 
-      Search_GPR_File (Root, Result);
+         --  If not found, perform a comprehensive search everywhere below
+         --  root. TODO: I'm not sure about that logic. As an user I might
```
reznikmm: I don't think comments in the code is right place to discuss server behavior. If you are not sure, open an issue instead of add a comment.

source/ada/lsp-ada_documents.adb
raph-amiard: I do not only rename it, I change its mode of operation: Before it operated on an Ada_Node, now it operates on a Decl, hence the renaming. Also, as you can see in this next commit:

https://github.com/AdaCore/ada_language_server/pull/22/commits/162c8f13d32e14acca7f98d6445dd3341fb404be

It's used as a basis to provide completion kind too. We can rename it Get_Decl_Symbol_Kind if you prefer, but ultimately it would be good to have all of those based on some function that operates on an abstract kind, in order to not duplicate the logic for every kind.

source/ada/lsp-ada_contexts.adb
raph-amiard: I agree, however a comment is better than nothing in the meantime ! I added a comment because I needed to understand what this function did.

source/ada/lsp-ada_driver.adb
```diff
@@ -58,5 +66,28 @@ begin
...
+      begin
+         Server.Run;
+      exception
+         when E : Property_Error =>
+            Server_Trace.Trace
+              ("LAL Property Error:" & Exception_Message (E));
+            Server_Trace.Trace (Symbolic_Traceback (E));
+
+         when Ada.IO_Exceptions.End_Error =>
+            Server_Trace.Trace
```
reznikmm: I would prefer Failure exit status here. One of the test in our tiny testsuite checks this.

raph-amiard: You mean rather than checking for End_Error ? This was the shortest path, I modified the function that raised a Constraint_Error before. We can further modify it to use  error codes instead, but I would suggest doing that in a second path.

source/ada/lsp-ada_contexts.adb
raph-amiard: I opened an issue and a reference to the issue in the comment. I prefer to keep the comment so we don't forget about it later :)

source/ada/lsp-ada_driver.adb
raph-amiard: Ah ok, I think I understood what you meant. You want the server to quit with an error code on this condition right ? In that case: OK, will take care of it !

FWIW: VS Code always quit like that on my system for some reason, which is why I didn't assume that it was an error case ...

source/ada/lsp-ada_documents.adb
reznikmm: I would keep these two kinds independent, so we will be able to change mapping from nodes to symbol kind without affecting mapping to completion kind.

source/ada/lsp-ada_contexts.adb
reznikmm: So, we have the issue and we won't forget it. Part after TODO: is not required any more.

source/ada/lsp-ada_documents.adb
raph-amiard: I disagree. 

1. I'm not going to duplicate code today because we might need that later. If we need different logic later, we'll take care of it later, it's simple enough.

2. There is no practical use case where you want a different kind for symbols and completions. Also keep in mind that all those kinds determine are icons.

source/ada/lsp-ada_contexts.adb
raph-amiard: Again, I disagree! It might not be apparent to somebody reading/modifying this code that there is something going on here that might need to be changed, and the information might be useful to her/him. Or, as @setton put it, there is very rarely too much comments. But whatever, will remove it, it's unimportant enough.

PR #29, opened 2018-11-26
Implement Process API using Glib

reznikmm:


ogorodnik:
@setton  I've already started porting this to GtkAda

reznikmm:
@setton Glib part has been moved to GtkAda.

PR #74, opened 2019-03-14
Topic/build gps

reznikmm:


.travis.yml
```diff
@@ -1,3 +1,5 @@
+dist: xenial
```
raph-amiard: This seems to be completely out of place, and not in line with PM and NicoSetton's comment on Slack. Why not let GPS's CI take care of this ?

raph-amiard: After discussion w. Anthony live: You guys discussed that this was a good compromise. Carry on :)

PR #75, opened 2019-03-20
S319-045 Revamp the test driver

setton:
Preliminary step to revamping the test drivers.
Introduce the new format, a test driver for this new format,
and a small utility to convert ALS traces to the new format.


testsuite/drivers/json_transformations.py
```diff
@@ -0,0 +1,128 @@
+""" A set of utilities for transforming json data for the purposes of
```
pmderodat: At this point, why not using YAML?

All JSON documents are valid YAML (and decoding them produce the same tree), the `yaml` module has the same interface as the `json` one, plus YAML allows comments.

pmderodat: Minor nit: I would suggest to reformat all docstring to comply to PEP 257 (as advertised in PEP8: https://www.python.org/dev/peps/pep-0008/#documentation-strings)

setton: Regarding YAML: there is a very good reason to prefer json in this particular case: the language server protocol is based on json messages... so, by using json, we can write input/out in the test case in a format that looks like the protocol itself, and we can easily import test cases from protocol traces.

Regarding PEP257: I'm not opposed, but I would like to discuss it live with the team, and will want to beef up our tooling before we do this - for instance have this being checked automatically in PRs.

raph-amiard: I agree about using the yaml parser, sounds like a good idea.

Regarding PEP257: I think it's weird do start here, given that few projects at AdaCore follow PEP257 (starting by Langkit/LAL)

pmderodat: You’re right about PEP257. Still, if we’re talking about what most projects at AdaCore do, the style we have here is not representative. So okay for not complying to the whole PEP257 right now, but at least reformat the docstrings in a more conventional way.

raph-amiard: :+1:

setton: I've added a patch to this that uses yaml in place of our own json-with-comments. Thans @pmderodat for the suggestion!

reznikmm: I don't agree. It would require extra work for writer (me?) to convert JSON request/responses from ALS/VSCode trace (which is JSON) to YAML. An extra place to make errors.

pmderodat: Thanks @setton!

@reznikmm But that’s exactly the point of using YAML: there is no conversion do in the first place. JSON documents are valid YAML and denote exactly the same structure.

PR #77, opened 2019-03-21
R430-030 Fix response parsing in the generic

reznikmm:


raph-amiard:
This kind of PR and review is not OK ...:

1. There is no comment about what this fixes
2. There is no test ?
3. The review & merge happened 10 minutes after the commit, leaving no chance to anyone else to review the commit.

We need to discuss again QA & review practices

setton:
I agree.

pmderodat:
Another thing surprises me: R430-030 was about the recent elaboration problem. Could you explain why do you use it here?

reznikmm:
Answering your questions:
1. I fixed parsing of responses in the generic. It's one line patch, so I guess this comment is enough.
1. No, we don't have tests for this part yet.
1. According to reviewing policy one reviewer is enough. The fix was urgent
1. I used the same ticket number that corresponds to fix introduced the bug. There isn't any other ticket for this problem.

pmderodat:
Thank you for your answers. I disagree with your point 1: this information probably would have been very useful to include in the commit message. This is what we should aim for, I think:

```
LSP.Generic_Responses: repair parsing

This change fixes the parsing of generic reponses, which was
broken by commit XXXX.

TN: R430-030
```

Also, what is currently in the way of writing a regression testcase for this? I mean, if we don’t have tests, maybe we should start to write some, no?

reznikmm:
We have a plan to test this part of code after GPS integration will be reviewed.

If you raise this topic, could you ask Raph where is tests for #37 #22?

pmderodat:
> We have a plan to test this part of code after GPS integration will be reviewed.

Ok. Personally I don’t think we should rely solely on integration tests to qualify software (this makes testing harder for contributors and generally gets in the way of good feature/code coverage), but that’s for another discussion.

I would still be interested in knowing how hard writing such a regression testcase in ALS would be, though. That could help the debate.

> If you raise this topic, could you ask Raph where is tests for #37 #22?

Seriously? Why haven’t you asked him yourself back, then? Or more constructively: no, the fact that we don’t have good standards does not mean that it justifies that we keep these low standards.

I really don’t like how this discussion is going. Can we please stay focus on what we should do? I see no problem on bringing up our previous mistake to the dialogue as long as it’s to learn from them, but it’s not acceptable to use them justify how badly we do things.

raph-amiard:
> If you raise this topic, could you ask Raph where is tests for #37 #22?

So, I completely agree, we should have tests for those, and I'm fully responsible for not writing them.

With that said: you were the one approving those PRs. You probably could have said something back then (or even after). This generally points towards the fact that we need to have a policy.

This points towards another problem: I tried to write some tests back then, and wasn't able to because I didn't understand how your infrastructure worked. The fact that it has no documentation, and that it doesn't follow common AdaCore standards with regards to how we build testsuites didn't help.

I should have asked for help but I did this work as part of a sprint in NYC with Raja and was then gone to the US tech days and working with Bob, so got sidetracked.

Anyway, the discussion is now happening, via Nico and Pierre-Marie, and you're basically refusing to make anything evolve. You will stay the only one writing tests if you refuse to listen to feedback about how it would be convenient to write tests, not only for you but for other contributors (which I am by the way)

PR #79, opened 2019-03-27
S327-016 Fix build failure on Windows

reznikmm:
This patch fixes build of Process API (Glib based implementation) on Windows.
When spawn related bindong for Glib was moved to GtkAda repository this file
wasn't fixed. Fix includes some renames to match code in GtkAda that is now
auto generated.


AnthonyLeonardoGracio:
Build works fine with that patch on my Windows ANCR setup

PR #80, opened 2019-03-29
Introduce two build modes: production and debug

pmderodat:


pmderodat:
Thanks for the quick review! :-)

PR #89, opened 2019-04-15
S328-046 Introduce asynchronous processing

setton:
Introduce a separate task for the actual processing, synchronized
to the main task via a queue of requests. This allows the main
task to never be blocking on input when libadalang is performing
an expensive computation. This will also be used to implement
a basic support for cancelling requests.

Also process the output in a separate task.

### DIFF
--- a/source/server/lsp-servers.ads
+++ b/source/server/lsp-servers.ads
@@ -116,9 +116,7 @@ private
    type Server is limited
      new LSP.Client_Notifications.Client_Notification_Handler with
    record
-      Initialized : Boolean;
-      Stop        : Boolean := False;
-      --  Mark Server as uninitialized until get 'initalize' request
+      Stop          : Boolean := False;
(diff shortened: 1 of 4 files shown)


source/server/lsp-servers.ads
```diff
@@ -50,24 +54,80 @@ package LSP.Servers is
...
+     Ada.Containers.Unbounded_Synchronized_Queues
+       (Unbounded_String_Queue_Interface);
+
+   type Requests_Queue_Access is access Requests_Queues.Queue;
+   type Output_Queue_Access is access Output_Queues.Queue;
+
+   --  The processing task
+   task type Processing_Task_Type is
+      entry Start
+        (In_Queue     : Requests_Queue_Access;
```
reznikmm: Have you considered an idea to make queue as discriminant of the type instead of passing it in the Initialize call?

PR #91, opened 2019-04-23
S405-010 Process defining_name in a different way

reznikmm:
Rewrite 'textDocument/definition' request:
 * if you click on identifier that not a defining_name (not a subprogram spec or body)
then return just single location of subprogram spec
 * if you click on defining_name of subprogram spec then return only single location of subprogram body
 * if you click on defining_name of subprogram body then return only single location of subprogram spec

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -124,16 +124,111 @@ package body LSP.Ada_Handlers is
       Value : LSP.Messages.TextDocumentPositionParams)
       return LSP.Messages.Location_Response
    is
+      use Libadalang.Analysis;
+
+      procedure Append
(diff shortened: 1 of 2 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -142,39 +237,29 @@ package body LSP.Ada_Handlers is
...
-      begin
-         Response.result.Append (Location);
-         return Response;
-      end;
+         if Other_Part /= No_Defining_Name then
+            Append (Response.result, Other_Part);
+         end if;
+      end if;
 
+      return Response;
```
setton: It would be great to have tests for the new behaviors

reznikmm: done

PR #96, opened 2019-05-03
(will abandon this) S502-015 switch notifications to the new decode workflow

setton:


source/ada/lsp-ada_handlers.adb
```diff
@@ -776,4 +796,31 @@ package body LSP.Ada_Handlers is
       return R;
    end Handle_Request;
 
+   -------------------------
+   -- Handle_Notification --
+   -------------------------
+
+   overriding procedure Handle_Notification
```
reznikmm: I would like message dispatching code to be in language independent part - in server, outside of ada/ directory. So each language server won't reimplement it.

source/protocol/lsp-server_notifications.ads
```diff
@@ -23,29 +23,8 @@ package LSP.Server_Notifications is
    type Server_Notification_Handler_Access is
      access all Server_Notification_Handler'Class;
```
reznikmm: Can we keep clean and precise interface for server instead of replacing it with just one procedure? Why not to use separate dispatcher interface with `Handle_Notification` method. The dispatcher then calls distinct method of handler. All servers will not reimplement dispatcher, but provide custom handler.

setton: Agreed. Will do this for both Requests and Notifications.

PR #97, opened 2019-05-09
S506-022 Add defensive code

setton:
Make sure any exception in finalization is fatal, preventing an
infinite loop on trying to exit.

Prevent any exception in task processing from quitting the task
main loops.

### DIFF
--- a/source/ada/lsp-ada_driver.adb
+++ b/source/ada/lsp-ada_driver.adb
@@ -33,6 +33,10 @@ with LSP.Ada_Handlers;
 
 with Libadalang.Common; use Libadalang.Common;
 
+--------------------
+-- LSP.Ada_Driver --
+--------------------
+
 procedure LSP.Ada_Driver is
 
    Server  : aliased LSP.Servers.Server;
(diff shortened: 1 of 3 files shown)


source/server/lsp-servers.adb
```diff
@@ -479,8 +482,22 @@ package body LSP.Servers is
                goto BEGIN_TASK_LOOP;
             end select;
 
-            --  Send the output to the stream
-            Write_JSON_RPC (Stream, Vector);
+            declare
```
godunko: Empty declaration block?

source/ada/lsp-ada_driver.adb
```diff
@@ -46,6 +50,7 @@ procedure LSP.Ada_Driver is
 
    ALS_Dir : constant Virtual_File := Get_Home_Directory / ".als";
    Do_Exit : Boolean := False;
+   Trying_To_Finalize : Boolean := False;
```
godunko: Do we definitely needs all this complexity? Why not to exit from the loop when server needs to be shutdown?

source/ada/lsp-ada_driver.adb
```diff
@@ -79,6 +84,7 @@ begin
 
       begin
          if Do_Exit then
+            Trying_To_Finalize := True;
             Server.Finalize;
             return;
```
godunko: Quite dangerous for me :) I suggest to move this code outside of the loop.

source/server/lsp-servers.adb
setton: yes, it's to capture the exception

source/ada/lsp-ada_driver.adb
setton: Fine by me, will make the change!

source/ada/lsp-ada_driver.adb
setton: done

PR #98, opened 2019-05-09
S506-022 make the URIs package accessible to clients

setton:
This is very convenient for clients to be able to convert from
file names to URIs without having to reimplement their routines.

Also change the test for checking that file names are absolute to
one that works with final directory separators.

### DIFF
--- a/source/uri/uris.ads
+++ b/source/uri/uris.ads
@@ -20,11 +20,11 @@
 --
 
 with Ada.Containers.Doubly_Linked_Lists;
-with Ada.Directories;
 with Ada.Iterator_Interfaces;
-with Ada.Strings.Fixed.Equal_Case_Insensitive;
 with Ada.Strings.Unbounded;
 with Ada.Strings.UTF_Encoding;
+
+with GNAT.OS_Lib;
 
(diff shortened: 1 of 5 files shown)


source/uri/uris.ads
```diff
@@ -95,10 +95,7 @@ package URIs is
 
    package Conversions is
       function From_File (Full_Path : String) return URI_String
-        with Pre =>
-          Ada.Strings.Fixed.Equal_Case_Insensitive
-            (Ada.Directories.Full_Name (Full_Path),
-             Full_Path);
+        with Pre => GNAT.OS_Lib.Is_Absolute_Path (Full_Path);
```
godunko: Good change but looks like unrelated to change "visibility" of URIs directory.

setton: Right, but I did make a note of it in the commit changelog.

setton:
> Also, did you checked that it doesn't break GPS build?

Yep!

source/uri/uris.ads
reznikmm: I'm not sure that implementation will still work with given precondition

PR #100, opened 2019-05-10
S329-011: Implement the textDocument/hover request

AnthonyLeonardoGracio:


source/ada/lsp-ada_handlers.adb
```diff
@@ -431,14 +434,206 @@ package body LSP.Ada_Handlers is
...
+              (Is_String => False,
+               value     => Hover_Text,
+               language  => To_LSP_String ("ada")));
+
+         --  TODO: append the trivias associated with the basic declaration
+         --  node when available in libadalang.
+         Response.result.contents.Append
+           (LSP.Messages.MarkedString'
+              (Is_String => True,
+               value     => To_LSP_String ("TODO: append the comments")));
```
setton: I wouldn't leave the "TODO" string visible to clients

source/ada/lsp-ada_handlers.adb
```diff
@@ -431,14 +434,206 @@ package body LSP.Ada_Handlers is
...
+                     Skip_Blanks (Lines (J).all, Idx);
+
+                     Blanks_Count_Per_Line (J) := Idx - Lines (J)'First;
+                     Indent_Blanks_Count := Natural'Min
+                       (Indent_Blanks_Count,
+                        Blanks_Count_Per_Line (J));
+                  end loop;
+
+                  for J in Lines'First + 1 .. Lines'Last loop
+                     declare
```
setton: (nit) I guess we could avoid this declare block and declaring Start_Idx..

source/ada/lsp-ada_handlers.adb
```diff
@@ -431,14 +434,206 @@ package body LSP.Ada_Handlers is
...
+              (Text,
+               On               => ASCII.LF,
+               Omit_Empty_Lines => True);
+            Idx         : Integer;
+         begin
+            --  If it's a single-line declaration, replace all the series of
+            --  whitespaces by only one blankspace.
+            --  If it's a multi-line declaration, remove only the unneeded
+            --  indentation whitespaces.
+
```
setton: for safety, should we add a case if Text = "" ? (it would only serve as defensive code, I don't think it can happen in real life)

source/ada/lsp-ada_handlers.adb
```diff
@@ -431,14 +434,206 @@ package body LSP.Ada_Handlers is
...
+                    (Lines'First + 1 .. Lines'Last) of Natural;
+                  Indent_Blanks_Count   : Natural := Natural'Last;
+               begin
+                  Hover_Text := To_LSP_String (Lines (Lines'First).all);
+
+                  --  Count the blankpaces per line and track how many
+                  --  blankspaces we should remove on each line by finding
+                  --  the common identation blankspaces.
+
+                  for J in Lines'First + 1 .. Lines'Last loop
```
setton: I may have missed something, but why do we not strip the leading blanks on the first line?

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: Well the trace that enables LSP based tooltips won't be enabled by default so they won't see it.

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: Ok will do

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: Ok will do

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: Because the Text function returns the node's text starting from its first token, without the trivias before (e.g: for '   procedure Something is null' it will return 'procedure Something is null', without the leading blankspaces).

PR #101, opened 2019-05-13
S506-022 Fix URIs.Conversions.From_File for full paths

setton:
The internal function Add_All_Paths doesn't expect
a directory which ends in a separator.

### DIFF
--- a/source/uri/uris.adb
+++ b/source/uri/uris.adb
@@ -65,6 +65,9 @@ package body URIs is
 
       function From_File (Full_Path : String) return URI_String is
 
+         Standardized_Full_Path : constant String :=
+           Ada.Directories.Full_Name (Full_Path);
+
          procedure Add_All_Paths
            (URI  : in out URIs.URI;
             Path : String;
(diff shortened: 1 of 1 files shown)


source/uri/uris.adb
```diff
@@ -104,7 +104,15 @@ package body URIs is
          --  Check if we have path in form of UNC:  \\host\share\path
 
          if Found (0) = GNAT.Regpat.No_Match then
-            Add_All_Paths (URI, Full_Path, "");
+            if Full_Path (Full_Path'Last)
```
reznikmm: I would use this way
```ada
Real_Full_Path : constant String := Ada.Directories.Full_Path (Full_Path);
begin
   return Real_From_File (Real_Full_Path);
end From_File;
```
And copy a From_File body to Real_From_File function.

This way we avoid using directory separator comparison. And this will work for several slashes at the end (like `/tmp/path///`).

setton: Good idea, done!

PR #107, opened 2019-05-15
S514-005 Publish diagnostics on file open

reznikmm:
Change textDocument/didOpen handler to publish diagnostics. Also
chage textDocument/didClose handler to clean diagnostics up.

### DIFF
--- a/source/ada/lsp-ada_contexts.ads
+++ b/source/ada/lsp-ada_contexts.ads
@@ -62,18 +62,21 @@ package LSP.Ada_Contexts is
      (Self : aliased in out Context;
       Item : LSP.Messages.TextDocumentItem)
       return LSP.Ada_Documents.Document_Access;
-   --  Load new document with text provided by LSP client.
+   --  Load new document with text provided by LSP client. Document is managed
+   --  by Context, user shouldn't free it or store it. Document will be
+   --  destroyed on Unload_Document call.
 
    procedure Unload_Document
      (Self : in out Context;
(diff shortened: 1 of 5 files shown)


source/ada/lsp-ada_contexts.ads
```diff
@@ -58,9 +58,10 @@ package LSP.Ada_Contexts is
    --  Reload the current context. This will invalidate and destroy any
    --  Libadalang related data, and recreate it from scratch.
 
-   procedure Load_Document
+   function Load_Document
      (Self : aliased in out Context;
-      Item : LSP.Messages.TextDocumentItem);
+      Item : LSP.Messages.TextDocumentItem)
+      return LSP.Ada_Documents.Document_Access;
```
setton: needs documentation on the return type, in particular whether or not the caller needs to free it (and how? should we call Unload_Document on it?)

source/ada/lsp-ada_handlers.adb
```diff
@@ -403,7 +409,12 @@ package body LSP.Ada_Handlers is
          end if;
       end if;
 
-      Self.Context.Load_Document (Value.textDocument);
+      Document := Self.Context.Load_Document (Value.textDocument);
+
+      Document.Get_Errors (Diag.diagnostics);
```
setton: (does Document need to be freed?)

reznikmm: No. All documents are managed by Context. I've added comments and disabled document deallocation outside of context.

PR #112, opened 2019-05-20
S516-036 avoid allocating large strings on the stack

setton:
Introduce a new function for string conversions
that avoids stack allocation, and use this in the
case of a giant string.

Adding test as part of the GPS testsuite.

### DIFF
--- a/source/protocol/lsp-types.adb
+++ b/source/protocol/lsp-types.adb
@@ -15,11 +15,16 @@
 -- of the license.                                                          --
 ------------------------------------------------------------------------------
 
+with Ada.Strings.Unbounded; use Ada.Strings.Unbounded;
+with Ada.Unchecked_Conversion;
 with Ada.Strings.UTF_Encoding.Wide_Strings;
 
 with LSP.JSON_Streams;
 
 package body LSP.Types is
+
+   Chunk_Size    : constant := 512;
+   --  When processing strings in chunks, this is the size of the chunk
 
    --------------
    -- Assigned --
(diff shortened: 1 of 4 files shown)


source/protocol/lsp-types.adb
```diff
@@ -166,6 +170,113 @@ package body LSP.Types is
...
+         --  Store one output code, C is in the range 0 .. 255
+
+         -----------
+         -- Store --
+         -----------
+
+         procedure Store (C : Unsigned_16) is
+         begin
+            Len := Len + 1;
+            Result (Len) := Character'Val (C);
```
reznikmm: Can we use this:
```
Ada.Strings.Unbounded.Append (Result, Character'Val (C));
```
Average complexity of Append is O(1), so not worse, but we avoid explicit `new` and `Free` in the code.

This way we can do even better: just iterate over Value taking slices of some reasonable length (say 512 characters) Encode them and append to Result. Like this:

```
Ada.Strings.Unbounded.Append (Result, Encode (Value (J .. J + 511)));
```

So we don't need copy/pasting of Encode.
What do you think?

setton: very good idea, I will try!

source/protocol/lsp-types.adb
```diff
@@ -166,6 +168,32 @@ package body LSP.Types is
...
+      Chunk_Size    : constant := 512;
+      Current_Index : Natural := Wide'First;
+      Next_Index    : Natural;
+   begin
+      --  Perform the encoding chunk by chunk, so as not to blow the stack
+      loop
+         Next_Index := Natural'Min (Current_Index + Chunk_Size, Wide'Last);
+         Ada.Strings.Unbounded.Append
+           (Res, Ada.Strings.UTF_Encoding.Wide_Strings.Encode
+              (Wide (Current_Index .. Next_Index)));
```
reznikmm: One more idea. We can avoid potentially large `Wide` string on the stack if replace `Wide (a .. b)` with `Slice` function from Wide_Unbounded_Strings package.

source/protocol/lsp-types.adb
setton: Done!

source/protocol/lsp-types.adb
setton: Done!

PR #116, opened 2019-05-22
Topic/no file fix

reznikmm:


raph-amiard:
Actually I personally don't think it's a big problem if a PR includes unrelated changes, as long as it is not too big, and that the fact is advertised in the title of the PR.

reznikmm:
This (testing codecs commit) is indeed unexpected. I'll redo PR

PR #119, opened 2019-05-28
S528-016 deactivate diagnostics by default

setton:
Create a trace, turned off by default, which controls the
emission of diagnostics.

To support testing diagnostics even if this feature is
swiched off:
  - add a mechanism which accepts a ".gnatdebug" in the current
    dir to specify traces, overriding ~/.als/traces.cfg
  - add a .gnatdebug activating the diagnostics trace in the
    test ada_lsp/publish_diag
  - modify the Makefile to run each test in its directory,
    so that there can be local .gnatdebug files for each tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 # Server executable file:
-export ALS=.obj/server/ada_language_server
+export ALS=$(shell pwd)/.obj/server/ada_language_server
 
 # Tester files
-TESTER=.obj/tester/tester-run
+TESTER=$(shell pwd)/.obj/tester/tester-run
 CODEC_TEST=.obj/codec_test/codec_test
 
 # Testsuite directory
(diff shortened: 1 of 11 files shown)


raph-amiard:
I'm really not fond of using traces in the ALS to activate/deactivate features. 

1.  First [opinion] because traces are not meant for that IMHO, and it's just a weird GPS idiom to use the trace file to also be a config file, rather than you know, use a config file

2. Second, because it will be unintuitive to reactivate. For me you should have a default (here probably deactivate diagnostics) that you can reactivate via a setting in the configChanged request.

Ideally, after a few small fixes to diagnostics, that I have locally, I would even prefer that diagnostics are on by default but deactivated from the GPS side, because the decision to not have diagnostics is more of a GPS one than an ALS one (eg. they are pretty usable in VS Code IMO)

setton:
> I'm really not fond of using traces in the ALS to activate/deactivate features.
> 
> 1. First [opinion] because traces are not meant for that IMHO, and it's just a weird GPS idiom to use the trace file to also be a config file, rather than you know, use a config file
> 2. Second, because it will be unintuitive to reactivate. For me you should have a default (here probably deactivate diagnostics) that you can reactivate via a setting in the configChanged request.


OK, I will add support for the configuration request. I will keep the code to activate "local traces" active, since this can be very useful, in particular for developers.

> Ideally, after a few small fixes to diagnostics, that I have locally, I would even prefer that diagnostics are on by default but deactivated from the GPS side, because the decision to not have diagnostics is more of a GPS one than an ALS one (eg. they are pretty usable in VS Code IMO)

Right.

PR #125, opened 2019-05-31
S531-034 Fix the conversion of URI to file

setton:
Avoids a potential constraint_error depending on where
the ALS was launched from, and simplify the URI conversion
function.

### DIFF
--- a/source/uri/uris.adb
+++ b/source/uri/uris.adb
@@ -28,37 +28,6 @@ package body URIs is
       UNC : constant GNAT.Regpat.Pattern_Matcher :=  --  \\host\share\path
         GNAT.Regpat.Compile ("^\\\\([^\\]*)\\");
 
-      function Find_Root return Ada.Strings.Unbounded.Unbounded_String;
-      --  Return "" on Windows and "/" otherwise
-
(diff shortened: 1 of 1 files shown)


setton:
> I spent a lot of time searching for working solution here. I doubt that the function could be simpler :)
> I propose to create tests for this. I'm afraid we can break already tested cases.

reznikmm:
I prefer to stay with implementation based on `Ada.Directories`. Windows path forms are rather complicated. Consider [File path formats on Windows systems](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats)

setton:
> I prefer to stay with implementation based on `Ada.Directories`. Windows path forms are rather complicated. Consider [File path formats on Windows systems](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats)

I don't understand at all why.

This is a function that converts from URI to file. So we don't have to support any of the various ways for *Windows* to render paths, we just have to support the way the URIs encode path, and then choose *one* way to render a path and emit this.

Could you clarify what the very complex recursive function brings that this patch doesn't also bring?

PR #128, opened 2019-06-03
Revert "S531-034 Fix the conversion of URI to file"

reznikmm:
Reverts AdaCore/ada_language_server#125.
Original issue is Ada.Directories bug. I propose to revert to original implementation.


setton:
my understanding is that the behavior of `Ada.Directories.Containing_Directory` has indeed changed, and that the new code is desirable...

PR #130, opened 2019-06-04
S603-029  implement "is called by" as a custom request

setton:
Introduce a custom request "textDocument/ALS_called_by".

Add a subprogram in LSP.LAL_Utils to implement the functionality,
add support for marshalling/unmarshalling the response to the
JSON stream, and add support for serving this request in the LSP
server.

Move the utility function Find_All_References to LSP.LAL_Utils.

Add a test for the 'called_by' functionality.

### DIFF
--- a/doc/called_by.md
+++ b/doc/called_by.md
@@ -1,0 +1,35 @@
+# Reference kinds
+
+## Short introduction
+
+This implements the functionality 'is called by', allowing to query
+all the calls to a function, organized by the entities where these
+calls happen.
+
+## Change description
+
+We introduce a new type to represent the results:
+
+```typescript
+
+interface ALSSubprogramAndReferences {
+
+   loc: Location; /* The location of the entity containing the references */
+   name: string; /* The name of the entity containing the references */
+
+   refs: Location[];  /* The references contained in this entity */
+}
+```
+
+And a new request:
+
+  method: `textDocument/ALS_called_by`
+  params: `TextDocumentPositionParams`
+
+Returning the references to the method identified at the given position:
+
+  result: `ALSSubprogramAndReferences[]`
+
+We also introduce a new boolean field `ALS_calledbyProvider` in the
+interface `ServerCapabilities` indicating whether the server supports
+this extension.
--- a/doc/called_by.md
+++ b/doc/called_by.md
@@ -6,6 +6,11 @@ This implements the functionality 'is called by', allowing to query
 all the calls to a function, organized by the entities where these
 calls happen.
 
+## Capabilities
+
+The `initialize` request returns a boolean `alsCalledByProvider` as part of
+the `capabilities`, set to true if the server supports this functionality.
+
 ## Change description
 
 We introduce a new type to represent the results:
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -113,6 +113,8 @@ package body LSP.Ada_Handlers is
           triggerCharacters => Empty_Vector & To_LSP_String (".")));
       Response.result.capabilities.hoverProvider := True;
 
+      Response.result.capabilities.ALS_calledbyProvider := True;
+
       if not LSP.Types.Is_Empty (Value.rootUri) then
          Root := Self.Context.URI_To_File (Value.rootUri);
       else
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -113,7 +113,7 @@ package body LSP.Ada_Handlers is
           triggerCharacters => Empty_Vector & To_LSP_String (".")));
       Response.result.capabilities.hoverProvider := True;
 
-      Response.result.capabilities.ALS_calledbyProvider := True;
+      Response.result.capabilities.alsCalledByProvider := True;
 
       if not LSP.Types.Is_Empty (Value.rootUri) then
          Root := Self.Context.URI_To_File (Value.rootUri);
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -769,41 +769,6 @@ package body LSP.Ada_Handlers is
    is
       use Libadalang.Analysis;
 
-      function Find_All_References
-        (Definition         : Defining_Name;
-         Sources            : GNATCOLL.VFS.File_Array_Access;
-         Include_Definition : Boolean := False)
-      return Ada_Node_Array;
-      --  Helper function, finds all references of a given defining name in a
-      --  given list of units.
-
-      function Find_All_References
-        (Definition         : Defining_Name;
-         Sources            : GNATCOLL.VFS.File_Array_Access;
-         Include_Definition : Boolean := False)
-      return Ada_Node_Array
-      is
-         Context : constant Analysis_Context := Definition.Unit.Context;
-         Source_Units : Analysis_Unit_Array (Sources'Range);
-      begin
-         for N in Sources'Range loop
-            Source_Units (N) := Context.Get_From_File
-              (Sources (N).Display_Full_Name,
-               Charset => Self.Context.Get_Charset);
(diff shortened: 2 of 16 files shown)


source/ada/lsp-lal_utils.adb
```diff
@@ -66,4 +70,132 @@ package body LSP.Lal_Utils is
...
+         Refs : constant Ada_Node_Array := Find_All_References
+           (Definition, Sources, Charset, Include_Definition => False);
+         Containing : Defining_Name;
+      begin
+         --  Go through all references to Name, organising them by containing
+         --  subprogram.
+
+         for Ref of Refs loop
+            --  Only consider references that are calls.
+            --  ??? To be discussed: how do we want to handle an access to
```
raph-amiard: Well it's not a direct call, and I'm pretty sure it won't be reported by the GNAT Xrefs, but indeed that could be interesting information.

I had a similar thought the other day about calls that go through generic instantiations, such as:

```ada
procedure Test is

   procedure A_Procedure is null;

   generic
      with procedure Foo;
   package A is
      procedure Bar;
   end A;

   package body A is
      procedure Bar is
      begin
         Foo;
      end Bar;
   end A;


   package A_Inst is new A (A_Procedure);

begin
   A_Procedure;
end Test;
```

Here, the call to `Foo` in the instantiation of `A_Inst` is most certainly a call to `A_Procedure`, yet neither GNAT nor LAL do find this at the moment, which is (IMHO) a bug.

Accesses are a more complicated case, because the reference is dynamic, of a similar problem I guess.

source/ada/lsp-lal_utils.adb
```diff
@@ -66,4 +70,132 @@ package body LSP.Lal_Utils is
...
+            --  a subprogram?
+            if Ref.As_Name.P_Is_Call then
+               --  We have a reference, and this a call: find the containing
+               --  subprogram or task
+               Containing := Containing_Entity (Ref);
+
+               if Containing /= No_Defining_Name then
+                  --  ??? Question: is it possible for the containing
```
raph-amiard: What is the question ? I don't understand

setton:
> It would be nice to fill protocol extension document for this extension as [described in doc/README.md](https://github.com/AdaCore/ada_language_server/tree/master/doc).
> 
> Also I propose to extend ServerCapabilities to let client know if server support this request.

Agreed, will do.

source/ada/lsp-ada_handlers.adb
```diff
@@ -43,6 +43,42 @@ package body LSP.Ada_Handlers is
                  return LSP.Types.LSP_String renames
      LSP.Types.To_LSP_String;
 
+   function To_Location
```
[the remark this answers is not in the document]
setton: I guess so, I can have a look.

source/ada/lsp-lal_utils.adb
```diff
@@ -66,4 +70,132 @@ package body LSP.Lal_Utils is
...
+            return References;
+         end if;
+      end;
+   end Find_All_References;
+
+   -----------------------
+   -- Containing_Entity --
+   -----------------------
+
+   function Containing_Entity (Ref : Ada_Node) return Defining_Name is
```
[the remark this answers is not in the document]
setton: OK! in the meantime we'll leave it here, so that we're not under pressure to bump the LAL branch etc.

source/ada/lsp-lal_utils.adb
setton: I agree that we should also provide an answer in your case.

I guess the high-level question is, when an user does an "is_called_by" request, what references are they interested in listing?

My current thinking is that they'd want not see everything - if the subprogram is being passed by access or through generic instantiations, it's better to list simply the point where the 'access is taken or the generics is instantiated than to omit it. Ie, I would consider all references except labels.

source/ada/lsp-lal_utils.adb
setton: thinko. Will remove.

source/ada/lsp-ada_handlers.adb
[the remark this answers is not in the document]
setton: had a look: first we should make File_To_URI independent of the context: I'll do this small revamp as a separate PR.

source/ada/lsp-lal_utils.adb
raph-amiard: I think we should go further actually ... :)

For generic instantiations, I'd even go as far as saying that those should be taken into account by find_all_refs (as an option). Not just the actual parameter to the generic instantiation, but like:

> Reference to A_Procedure [as A.Foo through generic instantiation A_Inst]

For pointers, probably find_all_refs shouldn't take them into account, but is_called_by should. For example, in the following example:

```ada
procedure Main is
   type Acc is access procedure;
   procedure Foo is null;
   procedure Bar is null;
   Inst : Acc;
begin
   Inst := Foo'Access;
   Inst.all;
end Main;
```

is_called_by should report something like:

> Possible call to Foo through access to subprogram

The logic would be extremely simple, and would only report a "possible call" for every access to subprograms, reporting the same in the following case:

```ada
procedure Main is
   type Acc is access procedure;
   procedure Foo is null;
   procedure Bar is null;
   Inst : Acc;
begin
   if Some_Condition then
      Inst := Foo'Access;
   else
      Inst := Bar'Access;
   end if;
   Inst.all;
end Main;
```

(this would also report a possible call to `Bar`).

The logic behind this thinking is that when you do "is_called_by", you actually want all places where the subp *might* be called. 

Which also brings us to dispatching subprograms, that are probably not handled by your code correctly right now. Given a base tagged type `Root` with a primitive procedure `Foo`, and two derived types `A` and `B` with overriding implementations for `Foo`, a call to `Root'Class.Foo` should also be taken into account as a possible call to both `A.Foo` and `B.Foo`.

To conclude (and I might/should have started with that), I think this logic should be part of LAL, not the LSP. LAL xref API should have an `is_called_by` function (and I think that's what we decided back in the day during our first meeting, but it wasn't written anywhere). So I think all of that is a job for @Roldak and you should just replace your code with a code to his implem of `is_called_by`.

doc/called_by.md
```diff
@@ -0,0 +1,35 @@
+# Reference kinds
+
+## Short introduction
+
+This implements the functionality 'is called by', allowing to query
+all the calls to a function, organized by the entities where these
+calls happen.
```
reznikmm: Absent description of new server capability.

doc/called_by.md
```diff
@@ -0,0 +1,35 @@
...
+   loc: Location; /* The location of the entity containing the references */
+   name: string; /* The name of the entity containing the references */
+
+   refs: Location[];  /* The references contained in this entity */
+}
+```
+
+And a new request:
+
+  method: `textDocument/ALS_called_by`
```
reznikmm: Shall we follow "camelCase" naming on method names here? See, for example `textDocument/onTypeFormatting`.

I propose `textDocument/alsCalledBy`. What do you think?

testsuite/ada_lsp/called_by/called_by.json
```diff
@@ -0,0 +1,181 @@
+[
+   {
+      "comment": [
+         "test automatically generated"
```
reznikmm: Can we have more meaningful comment here?

testsuite/ada_lsp/called_by/called_by.json
```diff
@@ -0,0 +1,181 @@
...
+            "id": 1, 
+            "method": "initialize"
+         }, 
+         "wait": [{
+               "id": 1, 
+               "result": {
+                  "capabilities": {
+                     "hoverProvider": true, 
+                     "documentSymbolProvider": true, 
+                     "referencesProvider": true,
```
reznikmm: Let's verify, that server sends new capability here

doc/called_by.md
raph-amiard: :+1: for me

PR #136, opened 2019-06-11
Misc work on exception reporting/better go to def results/diagnostics

raph-amiard:


source/ada/lsp-lal_utils.adb
```diff
@@ -54,7 +54,12 @@ package body LSP.Lal_Utils is
    ------------------
 
    function Resolve_Name (Name_Node : Name) return Defining_Name is
-      Result : constant Defining_Name := Name_Node.P_Xref;
+      --  In every case, if the unit has syntax errors, activate
+      --  Imprecise_Fallback.
+      --  TODO: Should we activate it in any case ?
```
AnthonyLeonardoGracio: I guess that all the ALS xref queries are done through this function no? In which other case should we activate this imprecise fallback mechanism?

raph-amiard: Yes, notably it should probably be activated for find_all_refs/other cases where we do xrefs.

It's a WIP :) Notably I still have to discuss with the LAL team, and you guys, whether we want it always on or not. My idea is that:

1. It should be always on, because it will only be used when either there is a nameres bug in LAL (which become more and more rare), or an error in the code. In case of error in the code, we want it on.

2. User should be notified when a result has been retrieved through the imprecise callback. There is some work on the LAL side to return xref results with the information (is this an imprecise result or not), but then I don't know how to make the *client* aware of that, and that depends on the ALS/LSP client too.

PR #137, opened 2019-06-14
S528-037 Recompute project view on didOpen unknown file.

reznikmm:
We also track list of opened files that doesn't belong to the project.
This allows correct work of cross-references for these files.

### DIFF
--- a/source/ada/lsp-ada_contexts.adb
+++ b/source/ada/lsp-ada_contexts.adb
@@ -179,7 +179,22 @@ package body LSP.Ada_Contexts is
       Root : LSP.Types.LSP_String) is
    begin
       Self.Root := Root;
+      Self.Source_Files := new File_Array'(1 .. 0 => <>);
    end Initialize;
+
+   ------------------------
+   -- Is_Part_Of_Project --
+   ------------------------
+
+   function Is_Part_Of_Project
+     (Self : Context;
+      File : Virtual_File) return Boolean
+   is
+      Set   : constant File_Info_Set := Self.Project_Tree.Info_Set (File);
+      First : constant File_Info'Class := File_Info'Class (Set.First_Element);
+   begin
+      return First.Project /= No_Project;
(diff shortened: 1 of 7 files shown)


source/ada/lsp-ada_contexts.adb
```diff
@@ -339,18 +368,69 @@ package body LSP.Ada_Contexts is
...
       Self.Documents.Delete (Item.uri);
       Free (Document);
+
+      declare
+         Name : constant LSP.Types.LSP_String := Self.URI_To_File (Item.uri);
+         File : constant Virtual_File := Create
+           (Filesystem_String (LSP.Types.To_UTF_8_String (Name)),
+            Normalize => True);
+      begin
+         if Self.Extra_Files.Contains (File) then
```
AnthonyLeonardoGracio: Can you please a comment of what you are doing here? Generallu speaking I think that you should comment a bit more your code, not only the subprogram declarations

source/ada/lsp-ada_contexts.adb
```diff
@@ -196,6 +211,18 @@ package body LSP.Ada_Contexts is
       Object.Initialize (Self.LAL_Context, Item);
       Self.Documents.Insert (Item.uri, Object);
 
+      declare
+         Name : constant LSP.Types.LSP_String := Self.URI_To_File (Item.uri);
+         File : constant Virtual_File := Create
+           (Filesystem_String (LSP.Types.To_UTF_8_String (Name)),
+            Normalize => True);
+      begin
+         if not Self.Is_Part_Of_Project (File) then
```
AnthonyLeonardoGracio: Please add a comment explaining why we are doing this. That might be obvious now but not 2, or 3 years later

source/ada/lsp-ada_contexts.adb
reznikmm: done

PR #139, opened 2019-06-20
S618-021 Support queries on non-open documents

setton:
Support the queries textdocument/hover and textdocument/alsCalledBy
on Uris for documents which are not currently open.

Add a test for the alsCalledBy request.

### DIFF
--- a/source/ada/lsp-ada_contexts.adb
+++ b/source/ada/lsp-ada_contexts.adb
@@ -148,6 +148,17 @@ package body LSP.Ada_Contexts is
    end Find_Project_File;
 
    ------------------
+   -- Has_Document --
+   ------------------
+
+   function Has_Document
+     (Self : Context;
+      URI  : LSP.Messages.DocumentUri) return Boolean is
+   begin
+      return Self.Documents.Contains (URI);
+   end Has_Document;
+
+   ------------------
    -- Get_Document --
    ------------------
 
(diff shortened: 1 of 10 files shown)


source/ada/lsp-ada_contexts.adb
```diff
@@ -510,4 +521,12 @@ package body LSP.Ada_Contexts is
       return Self.Diagnostics_Enabled;
    end Get_Diagnostics_Enabled;
 
+   ---------------------
+   -- Get_LAL_Context --
+   ---------------------
+
+   function Get_LAL_Context
```
reznikmm: why not extend Context interface instead of exposing Context properties?

source/ada/lsp-lal_utils.adb
```diff
@@ -201,4 +205,33 @@ package body LSP.Lal_Utils is
       return Result;
    end Is_Called_By;
 
+   -----------------
+   -- Get_Node_At --
+   -----------------
+
+   function Get_Node_At
```
reznikmm: I propose to move this function to lsp-ada_contexts.ads

source/ada/lsp-lal_utils.ads
```diff
@@ -43,6 +46,21 @@ package LSP.Lal_Utils is
    --  files from the disk.
    --  If Include_Definition is True, include the definition as well.
 
+   --------------------------
+   -- Document abstraction --
```
reznikmm: We already have document abstraction in LSP_Documents. I doubt we need another one.

reznikmm:
Ticket S618-021 doen't relate to the change

PR #142, opened 2019-06-25
S329-013 Send alsKind in textDocument/references responses.

reznikmm:
Add an extension to LSP to return reference kinds in
responses to textDocument/references.

### DIFF
--- a/doc/reference_kinds.md
+++ b/doc/reference_kinds.md
@@ -4,7 +4,7 @@
 
 This feature allows the LSP server to provide reference kinds in
 results for the `textDocument/references` request. A reference can
-be 'read', 'write', 'call', etc.
+be 'write', 'static call', 'dispatching call', etc.
 
 ## Change description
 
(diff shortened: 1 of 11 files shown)


doc/reference_kinds.md
```diff
@@ -13,18 +13,18 @@ extra field to the `Location` type:
...
 ```typescript
 
-export type AlsReferenceKind = 'read' | 'write' | 'call';
+export type AlsReferenceKind = 'w' | 'c' | 'd';
 
 export namespace AlsReferenceKind {
-   export const Read  : AlsReferenceKind = 'read';
-   export const Write : AlsReferenceKind = 'write';
-   export const Call  : AlsReferenceKind = 'call';
+   export const Write            : AlsReferenceKind = 'w';
```
setton: to prepare for the future, maybe we could add support for "Label", and for "Read" (in Ada we can have both Read+Write for an object that's passed as "in out")

reznikmm: I don't see why we should do it in advance. For me it's better to keep implementation in sync with specification. Besides 'Label' and 'Read' we already have a lot of (useless) kinds, I doubt anyone uses it ever.

PR #148, opened 2019-07-04
Consistently use short-circuit boolean operators

pmderodat:
As per AdaCore's coding style, and to avoid logic errors: programmers
usually rely on Y not to be evaluated when X is False in (X AND Y),
likewise for OR operations.


pmderodat:
Thanks Anthony! :-)

PR #150, opened 2019-07-04
Add code coverage computation capabilities to the testsuite

pmderodat:
While there are currently still hacks to do on top of this to get GNATcoverage actually work on ALS (pending bug fixes in gnatcov), these patches are all that’s needed to do in the ALS testsuite in order to compute code coverage with it.


pmderodat:
Good points! I got confused and thought `-gnatwae` was a single switch while there are two. I’ll split them and do the renaming you suggest. Thanks!

pmderodat:
As discussed on Slack: the project extensions are a temporary measure: our plan is to enhance GPRbuild so that generating these projects is not necessary anymore. So we should be able to simplify this in the near future.

Thanks for the review!

PR #151, opened 2019-07-05
S702-042 Update all stored Analisys_Units

reznikmm:
after each update of Libadalang Context. Otherwise LSP_Document
references outdated Analisys_Unit and some requests fail.

### DIFF
--- a/source/ada/lsp-ada_contexts.adb
+++ b/source/ada/lsp-ada_contexts.adb
@@ -375,10 +375,7 @@ package body LSP.Ada_Contexts is
          Self.Unit_Provider.Set (Provider);
       end;
 
-      Self.LAL_Context := Libadalang.Analysis.Create_Context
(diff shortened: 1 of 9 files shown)


reznikmm:
@raph-amiard I'm not sure that this will help in this particular patch. ALS doesn't keep sources of edited files by itself now. So after reloading LAL context we need to provide edited sources to new lal context. This is what the patch does. It iterates over all opened files, gets their text and puts into new context. Dropping analysis units doesn't change this, as I see.

raph-amiard:
@reznikmm Sorry about the "request changes", it's probably not adapted to this case, because I'm commenting on something that is already here prior to your PR.

However, my point is that you should *never* store an analysis unit in the document. Instead, just store the URI, and have a Document.Get_Unit method that goes and fetch the unit when you need it.

This way you don't have to worry about invalidation.

PR #153, opened 2019-07-08
S624-042 Pre-populate the lexical env for units

setton:
In the hope to improve the response time for user requests,
populate the lexical env for units when the files are being
opened.

Send 'log' traces around the call to Load_Document, for
development purposes.

### DIFF
--- a/source/ada/lsp-ada_documents.adb
+++ b/source/ada/lsp-ada_documents.adb
@@ -281,6 +281,12 @@ package body LSP.Ada_Documents is
          Buffer   => LSP.Types.To_UTF_8_Unbounded_String (Item.text));
       Self.URI := Item.uri;
       Self.LAL := LAL;
+
+      --  After creating an analysis unit, populate the lexical env with it:
+      --  we do this to allow Libadalang to do some work in reaction to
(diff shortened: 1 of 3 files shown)


source/ada/lsp-ada_documents.adb
```diff
@@ -281,6 +281,12 @@ package body LSP.Ada_Documents is
          Buffer   => LSP.Types.To_UTF_8_Unbounded_String (Item.text));
       Self.URI := Item.uri;
       Self.LAL := LAL;
+
+      --  After creating an analysis unit, populate the lexical env with it:
+      --  we do this to allow Libadalang to do some work in reaction to
+      --  a file being open in the IDE, in the hope to speed up the response
```
raph-amiard: I'm a bit concerned by this (and the commit message): Why "in the hope of" ? You should be able to test whether this actually completes your goal or not before doing a pull request no ?

setton: I did verify manually and it does seem to help.

raph-amiard: OK. Why not remove "in the hope" completely then ?

PR #156, opened 2019-07-12
S610-031: send messages when using Imprecise_Fallback

AnthonyLeonardoGracio:
We now send an LSP message when using the libadalang's imprecise
fallback mechanism: an info message is send for textDocument/definition
and a simple log message is sent in other cases.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -39,6 +39,8 @@ with Libadalang.Analysis;
 with Libadalang.Common;
 with Libadalang.Doc_Utils;
 
+with URIs;
+
 package body LSP.Ada_Handlers is
 
    function "+" (Text : Ada.Strings.UTF_Encoding.UTF_8_String)
@@ -51,6 +53,15 @@ package body LSP.Ada_Handlers is
       Kind : LSP.Messages.AlsReferenceKind_Set := LSP.Messages.Empty_Set)
       return LSP.Messages.Location;
    --  Return the location of the given node
+
+   procedure Send_Imprecise_Xref_Message
+     (Self     : access Message_Handler;
+      URI      : LSP.Messages.DocumentUri;
+      Position : LSP.Messages.Position;
+      Msg_Type : LSP.Messages.MessageType);
+   --  Send a message of the given Msg_Type to the LSP client to warn the user
+   --  of a possible imprecise result while computing xrefs on the given
+   --  node.
 
    -----------------------
    -- Get_Node_Location --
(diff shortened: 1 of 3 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -87,6 +96,27 @@ package body LSP.Ada_Handlers is
...
+     (Self     : access Message_Handler;
+      URI      : LSP.Messages.DocumentUri;
+      Position : LSP.Messages.Position;
+      Msg_Type : LSP.Messages.MessageType) is
+   begin
+      Self.Server.Show_Message
+        ((Msg_Type,
+         "Imprecise fallback used to compute cross-references on entity at:"
+         & To_LSP_String
+           (ASCII.LF & "   "  & To_UTF_8_String (URI))
```
setton: (minor) how  about we just send the base name of the file here,  to make the message less verbose?

source/ada/lsp-lal_utils.adb
```diff
@@ -57,14 +57,33 @@ package body LSP.Lal_Utils is
...
+     (Name_Node : Name;
+      Imprecise : out Boolean) return Defining_Name
+   is
+      Result : Defining_Name;
    begin
+      --  Try to get the cross reference without enabling the imprecise
+      --  fallback first.
+      --  Property_Error exceptions can be raised so make sure to handle
+      --  them correctly.
+      declare
```
setton: you don't need 'declare'  here, just begin  :)

source/ada/lsp-lal_utils.adb
```diff
@@ -57,14 +57,33 @@ package body LSP.Lal_Utils is
...
+      begin
+         Result := Name_Node.P_Xref (Imprecise_Fallback => True);
```
setton: shouldn't  we try first with  Imprecise_Fallback = False ?

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: True, let's do that

source/ada/lsp-lal_utils.adb
AnthonyLeonardoGracio: Indeed, I did not know that! (shame on me)

source/ada/lsp-lal_utils.adb
AnthonyLeonardoGracio: Wow, that was the purpose of my patch, thanks for noticing it

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: Actually I don't know how to convert the URI to a base name in a proper way... Do you know how to do it by chance?

source/ada/lsp-lal_utils.adb
setton: and I didn't know it two months ago :-)

source/ada/lsp-ada_handlers.adb
setton: I would use URIs.File_Conversions to convert to a VFS.Virtual_File first, then use Display_Base_Name

PR #164, opened 2019-07-28
Fix storing an unallocated variable

setton:


reznikmm:
I propose f81c48168336162daac6bb5d9c89fb63ec8c9047 instead

setton:
ok!

PR #166, opened 2019-07-28
S728-002 Do not store an Analysis_Unit in the Document

setton:
Analysis Units are not meant to be stored. Instead, store
the latest known version of the buffer, encoded in UTF-8, and
compute the Analysis Unit on demand.

### DIFF
--- a/source/ada/lsp-ada_documents.adb
+++ b/source/ada/lsp-ada_documents.adb
@@ -72,12 +72,13 @@ package body LSP.Ada_Documents is
       File : constant String :=
         Types.To_UTF_8_String
           (Self.Context.URI_To_File (Self.URI));  --  Delete file://
+      Dummy : Libadalang.Analysis.Analysis_Unit;
    begin
       Server_Trace.Trace ("Applying changes for document " & File);
       for Change of reverse Vector loop
          --  If whole document then reparse it
          if not Change.span.Is_Set then
-            Self.Unit := Self.LAL.Get_From_Buffer
+            Dummy := Self.LAL.Get_From_Buffer
               (Filename => File,
                --  Change.text is always encoded in UTF-8, as per the protocol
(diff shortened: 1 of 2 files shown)


source/ada/lsp-ada_documents.adb
```diff
@@ -443,4 +424,20 @@ package body LSP.Ada_Documents is
...
+   ----------
+   -- Unit --
+   ----------
+
+   function Unit (Self : Document) return Libadalang.Analysis.Analysis_Unit is
+      File : constant LSP.Types.LSP_String :=
+        Self.Context.URI_To_File (Self.URI);
+      Unit : Libadalang.Analysis.Analysis_Unit;
+   begin
+      Unit := Self.LAL.Get_From_Buffer
```
reznikmm: Why do we need Self.Buffer? LibAdaLang keeps text by it-self if we have already parsed it. And we parse file on DidOpen request. We can fetch Analysis_Unit at any moment using Self.LAL.Get_From_File.

setton: I don't understand this comment: we need to store the buffer in response to textdocument/didChange - we can't Get_From_File for buffers that contain unsaved changes.

reznikmm: I mean, we don't need the buffer.

When client opens/change the document, it send us document text and we pass it to LAL providing file name as unique document identifier. At any request we can get Analysis_Unit from LAL. To do this we just call `Get_From_File (File_Name, Reparse => False)`.

> we can't Get_From_File for buffers that contain unsaved changes.

Yes, we can! We have provided the change to LAL at `didChange` processing, so we just call Get_From_File and LAL won't read the file, but return already parsed unit.

All we need to implement this ticket, just drop Unit field from Document record.

PR #181, opened 2019-08-20
S820-009 Fix detection of imprecise name resolving

reznikmm:
Don't report name resolution as imprecise if we already has
defining_name.
Don't report it neither if name resolution fails.

### DIFF
--- a/source/ada/lsp-lal_utils.adb
+++ b/source/ada/lsp-lal_utils.adb
@@ -71,6 +71,13 @@ package body LSP.Lal_Utils is
       --  them correctly.
       begin
          Result := Name_Node.P_Xref (Imprecise_Fallback => False);
+
+         if Result = No_Defining_Name and then Name_Node.P_Is_Defining then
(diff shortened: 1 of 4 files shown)


reznikmm:
May be the commit comment isn't clear enough. Before this patch we send imprecise notification incorrectly. Each resolution of a defining_name is marked as imprecise, because we check P_Is_Defining after launching second p_xref (imprecise => True).

Another case, when we send imprecise notification incorrectly is when the resolution fails at all (both precise and imprecise p_xref fail). In this call (null-) answer should be considered as precise.

This patch fixes both issues.

I don't see any preference of p_referenced_decl over p_xref in LAL documentation. And I don't understand how this replacement allow us to better fix original issues.

reznikmm:
Looking at the documentation of the function:
>    function P_Defining_Name (Node : Basic_Decl'Class) return Defining_Name;
   --  Get the name of this declaration. If this declaration has several names,
   --  it will return the first one.

First name in name list is definitely not what we want.

PR #183, opened 2019-08-20
S815-010 Do not report error messages when loading the default project

setton:
Add explanations for this.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -342,6 +342,18 @@ package body LSP.Ada_Handlers is
          Errors);
 
       if not LSP.Types.Is_Empty (Errors.message) then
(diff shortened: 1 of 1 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -341,9 +341,15 @@ package body LSP.Ada_Handlers is
          "iso-8859-1",
          Errors);
 
-      if not LSP.Types.Is_Empty (Errors.message) then
-         Self.Server.On_Show_Message (Errors);
-      end if;
+      --  We might have encountered errors when loading the project, for
```
AnthonyLeonardoGracio: Can't we log something instead of reporting the messages? It might help for customer support

setton: Good idea, done in the new push!

PR #184, opened 2019-08-21
Topic/drop flood

reznikmm:


source/ada/lsp-ada_handlers.adb
```diff
@@ -642,10 +643,50 @@ package body LSP.Ada_Handlers is
      (Self  : access Message_Handler;
       Value : LSP.Messages.DidChangeTextDocumentParams)
    is
+      function Skip_Did_Change return Boolean;
+      --  Check if next message if didChange for the same document
```
setton: "Check if the following message in the queue is didChange for the same document"

PR #185, opened 2019-09-02
S819-026: Import renaming-related tests

AnthonyLeonardoGracio:


AnthonyLeonardoGracio:
Ok I will add all the missing yaml files in separate commits, completely forgot about that... The other imported tests don't have yaml files too.

PR #186, opened 2019-09-03
S701-035 Change format for list of reference kinds.

godunko:


source/protocol/lsp-messages.adb
```diff
@@ -105,11 +105,20 @@ package body LSP.Messages is
       UnknownErrorCode     => -32001,
       RequestCancelled     => -32800);
 
+   Write_Reference_Image            : aliased constant Standard.String :=
```
reznikmm: Could you fix doc/reference_kinds.md document too?

godunko:
I don't understand how test.yaml files are related to this patch :( There is no new tests here, only updates on existing one. If we miss these files and we need them for internal testing infrastructure we need to open internal ticket and handle it separately from this work.

PR #190, opened 2019-09-05
Remove the use of P_Xref

setton:
This call is obsolescent: replace its use in Resolve_Name
with the approach recommended by the Libadalang team.


source/ada/lsp-lal_utils.adb
```diff
@@ -65,34 +65,34 @@ package body LSP.Lal_Utils is
...
-         if Result = No_Defining_Name and then Name_Node.P_Is_Defining then
-            --  When Name_Node is part of defining_name and it isn't a
-            --  completion of another declaration, then P_Xref returns
-            --  No_Defining_Name. In this case we return current defining_name.
-            return Name_Node.P_Enclosing_Defining_Name;
+         if Name_Node.P_Is_Defining then
+            Result := Name_Node.P_Enclosing_Defining_Name.P_Canonical_Part;
+         else
+            Result := Name_Node.P_Referenced_Decl
+              (Imprecise_Fallback => False).P_Canonical_Part.P_Defining_Name;
```
reznikmm: As I've said last time, this won't work. See LAL documentation for P_Defining_Name:

>  p_defining_name
>     Get the name of this declaration. If this declaration has several names, it will return the first one.

This mean for this code
```#ada
declare
   A, B : Integer;
begin
  B := 0;  --  Try "Go to definition" here
end;
```

Go to definition brings you to "A" instead of B. This isn't what we want at all.

setton: @raph-amiard could you help here please?

Roldak: I'm not @raph-amiard but I think you should use `p_referenced_defining_name` now as a substitute for `p_xref`, which will bring you to the correct defining name of the declaration.

raph-amiard: Yes

setton: okay, I will wait for the next stable branch bump and make the change once `p_referenced_defining_name` is available.

raph-amiard: > okay, I will wait for the next stable branch bump and make the change once p_referenced_defining_name is available.

You can't because we're removing `p_xref` in the same go. So we need to synchronize the branch bump and this patch!

setton: Done! The latest version of this PR uses `p_referenced_defining_name`. Can you review @reznikmm @Roldak @raph-amiard ?

PR #199, opened 2019-09-18
S918-055 improve exception handling and error reporting

setton:
Improve the user experience around "find all references" and
"called by" operations:

  - first attempt the operation in precise mode, falling back
    on imprecise mode in case of Property_Error

  - even in imprecise mode, catch Property_Error and send empty
    results

  - send a message to the user when imprecise results are being
    returned

  - in the case of 'rename', abort the rename operation if the results
    are not precise.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -91,6 +91,12 @@ package body LSP.Ada_Handlers is
      (Create (+(URIs.Conversions.To_File (To_UTF_8_String (URI)))));
    --  Utility conversion function
 
+   procedure Show_Message
+     (Self : access Message_Handler;
+      Text : String;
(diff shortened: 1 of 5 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -85,12 +85,18 @@ package body LSP.Ada_Handlers is
    --  Append given Node location to the Result.
    --  Do nothing if the item inside of an synthetic file (like __standard).
    --  Do not append if the location is already in Result.
-   --  See description of Kind in Get_Node_Location comments.
+   --  See desc/ription of Kind in Get_Node_Location comments.
```
reznikmm: An extra slash ('/')

setton: thank you! fixed

PR #201, opened 2019-09-20
S822-037 inform alsReferenceKind in the result of alsCalledBy

setton:
This allows the client to differentiate calls that are made
through dispatching, for instance.

Add test.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -1328,6 +1328,33 @@ package body LSP.Ada_Handlers is
         (Is_Error => False);
       Imprecise  : Boolean := False;
 
+      function Get_Reference_Kind
+        (Node : Name) return LSP.Messages.AlsReferenceKind_Set;
+      --  Query whether Node is a static or a dispatching call, and format
+      --  this into an AlsReferenceKind_Set.
+
+      ------------------------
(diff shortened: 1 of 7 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -1324,6 +1324,33 @@ package body LSP.Ada_Handlers is
...
+      is
+         Result : LSP.Messages.AlsReferenceKind_Set := LSP.Messages.Empty_Set;
+      begin
+         begin
+            Result.As_Flags (LSP.Messages.Static_Call) :=
+              Node.P_Is_Static_Call;
+            Result.As_Flags (LSP.Messages.Dispatching_Call) :=
+              Node.P_Is_Dispatching_Call;
+         exception
+            when Libadalang.Common.Property_Error =>
```
AdrienBoulanger: We should at least trace the exception

setton: I agree, will do!

setton: I will do it with the system put in place in https://github.com/AdaCore/ada_language_server/pull/204 once this is pushed.

setton: done!

PR #204, opened 2019-09-25
S925-011 log all "Property_Error" exceptions to traces

setton:
We're often expecting this exception and catching it so that
the user does not see it: capture these in the traces, to ease
debugging and investigation.

### DIFF
--- a/source/ada/lsp-common.adb
+++ b/source/ada/lsp-common.adb
@@ -1,0 +1,40 @@
+------------------------------------------------------------------------------
+--                         Language Server Protocol                         --
+--                                                                          --
+--                     Copyright (C) 2018-2019, AdaCore                     --
+--                                                                          --
+-- This is free software;  you can redistribute it  and/or modify it  under --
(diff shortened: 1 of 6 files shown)


source/ada/lsp-common.adb
```diff
@@ -0,0 +1,39 @@
...
+   procedure Log
+     (Trace   : GNATCOLL.Traces.Trace_Handle;
+      E       : Ada.Exceptions.Exception_Occurrence;
+      Message : String := "") is
+   begin
+      if Message /= "" then
+         Trace.Trace (Message);
+      end if;
+
+      Trace.Trace (Exception_Name (E) & ASCII.LF & Symbolic_Traceback (E));
```
AnthonyLeonardoGracio: The TN consists to log the message attached to the Property_Error exceptions, so we should call Exception_Message here too, unless I am missing something

setton: I think you're right, can't hurt in the general case! will do.

setton: done!

PR #208, opened 2019-10-06
SA02-039 Don't return error responses when Property_Error is raised

reznikmm:


source/ada/lsp-ada_driver.adb
```diff
@@ -50,6 +50,11 @@ procedure LSP.Ada_Driver is
    Handler : aliased LSP.Ada_Handlers.Message_Handler
      (Server'Access, Server_Trace);
 
+   Property_Error_Decorator : aliased LSP.Property_Error_Decorators
+     .Property_Error_Decorator (Server_Trace, Handler'Unchecked_Access);
+   --  This decorator catches all Property_Error exception and provides
```
setton: * exceptions

source/ada/lsp-property_error_decorators.adb
```diff
@@ -0,0 +1,332 @@
...
+
+   overriding function On_CodeAction_Request
+     (Self    : access Property_Error_Decorator;
+      Request : LSP.Messages.Server_Requests.CodeAction_Request)
+      return LSP.Messages.Server_Responses.CodeAction_Response is
+   begin
+      return Self.Handler.On_CodeAction_Request (Request);
+   exception
+      when E : Libadalang.Common.Property_Error =>
+         Self.Trace.Trace ("Uncatched Property_Error");
```
setton: * Uncatched -> Uncaught

(many occurrences)

reznikmm:
My idea was that if we have a separate module with a single purpose to catch Property_Error in every LSP request, that we will have less chance to forget adding corresponding exception clause in a particular request procedure.

What do you think?

I can try to use generic procedure to avoid code duplication. it will looks like:

```
   procedure CodeAction_Request is new Generic_Request
     (LSP.Messages.Server_Requests.CodeAction_Request,
      LSP.Messages.Server_Responses.CodeAction_Response);

   overriding function On_CodeAction_Request
     (Self    : access Property_Error_Decorator;
      Request : LSP.Messages.Server_Requests.CodeAction_Request)
      return LSP.Messages.Server_Responses.CodeAction_Response
       renames CodeAction_Request

```

setton:
I think the general idea is good but I propose a small twist: rather than have a specific `Property_Handle_Decorator` to handle just a property_error, we add a global exception handler in the task that processes requests. And then, to be able to answer something to the client in case of an exception, we create a type `type Fallback_Handler (...) is new LSP.Server_Request_Handlers.Server_Request_Handler` which will simply return an empty response of the appropriate type for every request. If you agree I'm happy to do this in a separate commit.

raph-amiard:
> I think the general idea is good but I propose a small twist: rather than have a specific Property_Handle_Decorator to handle just a property_error, we add a global exception handler in the task that processes requests. And then, to be able to answer something to the client in case of an exception, we create a type type Fallback_Handler (...) is new LSP.Server_Request_Handlers.Server_Request_Handler which will simply return an empty response of the appropriate type for every request. If you agree I'm happy to do this in a separate commit.

I'm confused, didn't we have this already?

raph-amiard:
Ok I guess I understand: for the moment we have a catch-all, but it still sends an exception response:

https://github.com/AdaCore/ada_language_server/blob/master/source/server/lsp-servers.adb#L946

@reznikmm from my point of view, it seems like what you're doing here could be maybe done in a simpler way at this point where we send an exception response? But maybe I'm missing something?

reznikmm:
@raph-amiard Generally speaking each request has its-own response format. That's why I proposed per request default responses.

reznikmm:
I've dropped en extra code found by @raph-amiard and I've rewritten decorator using generics as @godunko suggested. I renamed Property_Error_Decorator to Error_Decorator and moved context reloading call after `when others =>` handler there.

PR #212, opened 2019-10-10
Document neovim integration

raph-amiard:


README.md
```diff
@@ -173,6 +173,42 @@ Here is an example config file from the gnatcov project:
...
+    \ 'ada': ['path/to/ada_language_server'],
+    \ }
+" if you already have LanguageClient_serverCommands, just add a line for ada.
+```
+
+To configure the ALS for a specific workspace/project, you can use the
+`.vim/settings.json` file. It is mandatory as soon as you want to use a
+specific `.gpr` project file.
+
+This is the way to specify a project file, eg. you cannot open a project file
```
setton: will ada_language_server be launched in the same directory as vim? If so, I guess we could say that, if there's a single .gpr present, then it will be used, otherwise a project should be explicitly given in .vim/settings.json with (...) ?

raph-amiard: I don't know (didn't check), but it isn't documented as such (ie. with what the server is spawned as cwd isn't documented), so I would be wary of documenting that.

raph-amiard:
@briot That's a fair point, and I thought about it myself

I added an explanatory paragraph, but I'm not sure how helpful it is... Don't hesitate to give me some more feedback about it.

README.md
```diff
@@ -173,6 +173,49 @@ Here is an example config file from the gnatcov project:
...
+    "ada.projectFile": "project.gpr",
+    "ada.scenarioVariables": {
+        "GLFW_Version": "3",
+        "GLFW_Lib": "-lglfw",
+        "Windowing_System": "x11"
+    }
+}
+```
+
+The location where the `.vim` folder is located will determine the relative
```
briot: This explanation is good enough now, thanks for the extra precision

README.md
briot: Looking at the Rust sources for the vim plugin, the language server is spawned in `src/language_server_protocol.rs:2956`. It seems to use a variable `rootPath` as the current directory. This variable is extracted from the parameters (which I assume can be set in the vim configuration, but I haven't found that yet). If none is provided, it is computed via the function `getRootPath` in `src/utils.rs`. The latter has a number of hard-coded algorithm to compute this path depending on the language.  For Ada, it would default to the directory containing `.git`, `.hg` or `.svn`

briot: See also https://github.com/neoclide/coc.nvim/issues/107

One interesting comment there is `It is configurable, use g:rooter_patterns, check out :h g:rooter_patterns, same as vim-rooter.` which maybe is also related to how the `.vim` directory is found.

briot:
For one of the remaining limitations, i.e. there is a single project defined in a json config file, I have come up with the following vimscript function. This lets you specify the name of a project file to use, will kill the currently running server, update settings.json, and restart the server. It seems to work fine for the time being. Feel free to include it in your patch if you want to.

```
" A function that changes the project file used by the language server.
" It restarts the language server.
" Provides command line completion for files.
function! DbcReloadLS(project)
   " Nothing to do if we haven't loaded the plugin
   if get(g:, "LanguageClient_loaded") == 0
      echo "LanguageClient plugin not loaded"
      return
   endif

   " Kill a running instance of the language server
   call LanguageClient#exit()

   " First compute the absolute path of the project. Opening the settings
   " file might modify the cwd if the user has set "autochdir"
   let l:project = fnamemodify(a:project, ":p")

   " Then open the settings file
   " ??? Should use the LSP variable to find the directory and name of the
   " settings file.

   let l:dir = finddir(".vim", ".;")
   let l:filename = l:dir . "/settings.json"
   execute "edit! " . l:filename

   " Remove the previous ada.projectFile line, and replace with our own
   call cursor(1, 1)
   let l:line = search("ada.projectFile")
   call append(l:line, '    "ada.projectFile": "' . l:project . '",')
   execute line 'delete _'
   execute "write!"

   " Close the temporary buffer for settings.json, but keep window arrangement
   " as it is
   execute "e#\|bd #"
   "execute "bdelete"

   " Restart the language server to load the new config
   " Without the delay, the server doesn't restart properly
   sleep 100m
   call LanguageClient#startServer()
endfunction
command -nargs=1 -complete=file DbcReloadLS  :call DbcReloadLS(<f-args>)
```

PR #218, opened 2019-10-23
SA11-040 Add a fallback mechanism for 'definition'

setton:
This mechanism is based on a naive heuristics and is meant
for the language server to be able to provide navigation from
a subprogram spec to the body (and vice versa) after modifying
the profile in one of the parts but not the other.

Add a test.

### DIFF
--- a/source/ada/lsp-lal_utils.adb
+++ b/source/ada/lsp-lal_utils.adb
@@ -18,6 +18,10 @@
 with LSP.Common;        use LSP.Common;
 with Libadalang.Common; use Libadalang.Common;
 
+with Langkit_Support;
+with Langkit_Support.Text;
+with Langkit_Support.Slocs;
+
 package body LSP.Lal_Utils is
 
    function Containing_Entity (Ref : Ada_Node) return Defining_Name;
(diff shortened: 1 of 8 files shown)


source/ada/lsp-lal_utils.adb
```diff
@@ -103,6 +107,142 @@ package body LSP.Lal_Utils is
...
+         Parent_Body := Find_Next_Part (Parent_Spec, Trace);
+         if Parent_Body = No_Defining_Name then
+            Parent_Body := Find_Next_Part (Parent_Spec, Trace);
+         end if;
+         if Parent_Body /= No_Defining_Name then
+            Parent_Body.Parent.Traverse (Matches'Unrestricted_Access);
+         end if;
+      end if;
+
+      return Found;
```
reznikmm: Let's catch Property_Error here.

source/ada/lsp-lal_utils.adb
```diff
@@ -103,6 +107,142 @@ package body LSP.Lal_Utils is
...
+   -- Find_First_Part --
+   ---------------------
+
+   function Find_Canonical_Part
+     (Definition : Defining_Name;
+      Trace      : GNATCOLL.Traces.Trace_Handle) return Defining_Name
+   is
+      Canonical : Defining_Name;
+   begin
+      --  Iterate over Next.P_Previous_Part names until no name found
```
reznikmm: it seems, comment is out of date

source/ada/lsp-lal_utils.adb
setton: Done!

source/ada/lsp-lal_utils.adb
setton: Indeed, removed.

PR #219, opened 2019-10-24
SA24-025 allow "called_by" on an abstract definition

setton:
Add a test.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -1386,6 +1386,7 @@ package body LSP.Ada_Handlers is
          if Definition = No_Defining_Name
            or else Definition.P_Basic_Decl.Kind not in
              Ada_Subp_Decl | Ada_Subp_Body | Ada_Null_Subp_Decl
+               | Ada_Abstract_Subp_Decl
                | Ada_Entry_Decl | Ada_Entry_Body
            or else Request.Canceled
          then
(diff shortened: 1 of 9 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -1386,6 +1386,7 @@ package body LSP.Ada_Handlers is
          if Definition = No_Defining_Name
            or else Definition.P_Basic_Decl.Kind not in
              Ada_Subp_Decl | Ada_Subp_Body | Ada_Null_Subp_Decl
+               | Ada_Abstract_Subp_Decl
                | Ada_Entry_Decl | Ada_Entry_Body
```
AnthonyLeonardoGracio: As discussed, we should use either an appropriate subprogram or LAL substypes for this kind of checks

source/ada/lsp-lal_utils.ads
```diff
@@ -66,6 +66,7 @@ package LSP.Lal_Utils is
      with Pre =>
        Definition.P_Basic_Decl.Kind in Libadalang.Common.Ada_Subp_Decl
          | Libadalang.Common.Ada_Subp_Body
+         | Libadalang.Common.Ada_Abstract_Subp_Decl
          | Libadalang.Common.Ada_Null_Subp_Decl
          | Libadalang.Common.Ada_Entry_Decl
```
AnthonyLeonardoGracio: Same thing here

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: We should try to use subprograms or subtypes to check this (as done in the previous commit).

source/ada/lsp-lal_utils.ads
```diff
@@ -66,6 +66,7 @@ package LSP.Lal_Utils is
...
          | Libadalang.Common.Ada_Entry_Body;
```
AnthonyLeonardoGracio: Same thing here

PR #222, opened 2019-10-29
SA21-052 Add progress reporting and indexing at project load

setton:


source/ada/lsp-ada_handlers.adb
```diff
@@ -1808,8 +1817,131 @@ package body LSP.Ada_Handlers is
...
+
+      Index           : Natural := 1;
+      Total           : Natural := 0;
+      Last_Percent    : Natural := 0;
+      Current_Percent : Natural := 0;
+   begin
+      --  Collect the contexts and their sources: we do this so that we are
+      --  able to produce a progress bar covering the total number of files.
+
+      for J of Self.Contexts loop
```
AnthonyLeonardoGracio: I would prefer to use a more explicit name here, like 'Context' (J is more used for integer-based indexes)

source/ada/lsp-ada_handlers.adb
```diff
@@ -1808,8 +1817,131 @@ package body LSP.Ada_Handlers is
...
+      type Context_And_Files is record
+         Context : Context_Access;
+         Sources : File_Array_Access;
+      end record;
+      type Context_And_Files_Array is array
+        (Natural range <>) of Context_And_Files;
+
+      Arr   : Context_And_Files_Array (1 .. Natural (Self.Contexts.Length));
+
+      token : constant LSP.Types.LSP_Number_Or_String
```
AnthonyLeonardoGracio: 'Token' instead of 'token'?

source/ada/lsp-ada_handlers.adb
setton: Will do!

source/ada/lsp-ada_handlers.adb
setton: For this I'm deliberately using the casing provided by the protocol, since this is a protocol object. But it's not at all a strong opinion :-)

AnthonyLeonardoGracio: Ah I see, fine by me then, keep the protocol casing here

source/ada/lsp-ada_handlers.adb
setton: done!

PR #224, opened 2019-11-13
[enh] Add Problem Matcher

dgautherot:
Hello,

I was about to create an Ada language support extension for Visual Studio Code when I discovered your repo and I thing you are on the right way by using the Language Server Protocol.

I would love to contribute to your work with this first pull request, adding a problem matcher for this extension.

Would it be possible to integrate it an push a new version of the Visual Studio Extension onto its marketplace ?

Thanks a lot.
Kind regard,

Denis Gautherot.

### DIFF
--- a/integration/vscode/ada/package.json
+++ b/integration/vscode/ada/package.json
@@ -99,6 +99,23 @@
                     }
                 }
             }
+        ],
+        "problemMatchers": [
+            {
+                "name": "ada",
+                "label": "Gnat Ada Problem Matcher",
+                "fileLocation": ["relative", "${workspaceRoot}"],
+                "pattern": [
+                    {
+                        "regexp": "^(.*):(\\d+):(\\d+):\\s+(warning|)(?:\\s+)?(.*)$",
+                        "file": 1,
+                        "line": 2,
+                        "column": 3,
+                        "severity": 4,
+                        "message": 5
+                    }
+                ]
+            }
         ]
     },
     "devDependencies": {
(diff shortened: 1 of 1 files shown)


setton:
Thank you @dgautherot! We'll take a look ASAP.

integration/vscode/ada/package.json
```diff
@@ -99,6 +99,23 @@
...
             }
+        ],
+        "problemMatchers": [
+            {
+                "name": "ada",
+                "label": "Gnat Ada Problem Matcher",
+                "fileLocation": ["relative", "${workspaceRoot}"],
+                "pattern": [
+                    {
+                        "regexp": "^(.*):(\\d+):(\\d+):\\s+(warning|)(?:\\s+)?(.*)$",
```
reznikmm: I'm afraid first part of the regexp won't work on Windows. (Windows has drive letters, like `C:\`). I propose replace `^(.*)` with something like this `^([^:]:?[^:]*)`

What do you think?

dgautherot: Hello reznikmm!
I'm fine with your proposal!
So here would be the final regex (i've tested with specific cases and noticed a little typo with the final ':' that was included into the message part):

"^([^:]:?[^:]*):(\\d+):(\\d+):\\s+(warning|)(?:[:]\\s+)?(.*)$"

reznikmm:
@dgautherot New version of the extension (v21.0.2) is on the marketplace! Thank you for contributing!

PR #238, opened 2019-11-18
Topic/update protocol

reznikmm:
Update LSP.Messages types to LSP 3.14 version.

### DIFF
--- a/source/protocol/lsp-generic_sets.ads
+++ b/source/protocol/lsp-generic_sets.ads
@@ -1,0 +1,64 @@
+------------------------------------------------------------------------------
+--                         Language Server Protocol                         --
+--                                                                          --
+--                     Copyright (C) 2018-2019, AdaCore                     --
+--                                                                          --
+-- This is free software;  you can redistribute it  and/or modify it  under --
+-- terms of the  GNU General Public License as published  by the Free Soft- --
+-- ware  Foundation;  either version 3,  or (at your option) any later ver- --
+-- sion.  This software is distributed in the hope  that it will be useful, --
+-- but WITHOUT ANY WARRANTY;  without even the implied warranty of MERCHAN- --
+-- TABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public --
+-- License for  more details.  You should have  received  a copy of the GNU --
+-- General  Public  License  distributed  with  this  software;   see  file --
+-- COPYING3.  If not, go to http://www.gnu.org/licenses for a complete copy --
+-- of the license.                                                          --
+------------------------------------------------------------------------------
+
+--  This package provides generic set to implement Language Server Protocol.
+
+with Ada.Streams;
+
+generic
+   type Element is (<>);
+
+package LSP.Generic_Sets is
+
+   type Set is private;
+
+   function Empty return Set
+     with Inline;
+   --  Return an empty set
+
+   function To_Set (From, To : Element) return Set
+     with Inline;
+   --  Return a set which values in range From .. To
+
+   function Contains (Self : Set; Value : Element) return Boolean
+     with Inline;
+   --  Check if Value is in the set
+
+   procedure Include (Self : in out Set; Value : Element)
+     with Inline;
+   --  Append Value to the set
+
+   procedure Remove (Self : in out Set; Value : Element)
+     with Inline;
+   --  Remove Value to the set
+
+private
+
+   type Set is array (Element) of Boolean;
+
+   procedure Read_Set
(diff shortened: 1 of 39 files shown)


doc/reorder.awk
```diff
@@ -0,0 +1,115 @@
...
+# Use next command to run the script:
+#
+#sed -n -e '/^   --```typescript/,/^   --```/p' \
+# ../source/protocol/lsp-messages.ads | awk -f reorder.awk
+BEGIN{Span_Index=0}
+/--```typescript/{Line_Index=0}
+!/--```/{ARR[Span_Index][Line_Index++]=$0}
+/--```$/{Span_Index++;Line_Index=0}
+END{
+    M=0
```
setton: this seems really hard to maintain - I'm not sure how it works... do we need to check this in?

doc/reorder.awk
```diff
@@ -0,0 +1,115 @@
+# This AWK script extracts typescripts comments from lsp-messages.ads
+# and reorders them in an order that corresponds to LSP specification.
+# Use next command to run the script:
```
setton: * Use the following command to run the script:

source/protocol/lsp-generic_sets.ads
```diff
@@ -0,0 +1,40 @@
...
+------------------------------------------------------------------------------
+
+--  This package provides generic set to implement Language Server Protocol.
+
+with Ada.Streams;
+
+generic
+   type Element is (<>);
+
+package LSP.Generic_Sets is
```
setton: `Generic_Sets` does not seem the right name for this - is this really a set? Could we call it `Generic_Boolean_Arrays` and change the type from `Set` to `Boolean_Array` ?

source/protocol/lsp-json_streams.ads
```diff
@@ -107,8 +105,9 @@ private
    package State_Vectors is new Ada.Containers.Vectors (Positive, State);
 
    type JSON_Stream is new Ada.Streams.Root_Stream_Type with record
-      Current : State;
-      Stack   : State_Vectors.Vector;
+      Writting : Boolean := True;  --  True means stream to write
```
setton: * `Writing` 
How about `Writable` ?

source/protocol/lsp-generic_sets.ads
reznikmm: It's a set of enumerated values. It's represented as array of Boolean. So you can check if a value in a set with `if Set (Value) then`. I guess, this is well-known way to have effective set of small number of values in Ada.

doc/reorder.awk
reznikmm: lsp-messages.ads contains typescript pieces in comments surrounded by kind of brackets:
```
--```typescript
--   ... typescript code here ...
--```
```
Typescript pieces are copied from LSP specification. The problem here is that we need to change their order, because Ada requires forward declarations, while LSP specification doesn't use this order. This script extracts typescripts pieces and prints them in original order. After that one can diff result with corresponding text from LSP specification.

I'm afraid to lost this information (relative order of typescript pieces), so until we have better way to do diff, I would prefer to keep this in repo.

source/protocol/lsp-generic_sets.ads
setton: What I'm trying to do is avoid the confusion with `Set` as in the container (`Hashed_Set` and so forth). How about `Generic_Bitfield` or (by favourite) `Generic_Flags` for the package name? And call the function `Is_Set`, so clients can call `if Is_Set (Value) then`.

reznikmm: But this Set **is** a container. There isn't any semantic difference between `new Generic_Sets (My_Enum)` and `new Ada.Container.Hashed_Sets (My_Enum)`. I just provide more effective implementation.

Ok. I'll hide the array in the private part and put standard `Contains`, `Include`, `Delete` subprograms into visible part, if you want.

source/protocol/lsp-json_streams.ads
reznikmm: :heavy_check_mark: done!

doc/reorder.awk
reznikmm: :heavy_check_mark: done!

source/protocol/lsp-generic_sets.ads
setton: no - if your implementation is the same semantic as `Ada.Containers.Hashed_Sets`, then we should use an `Ada.Containers.Hashed_Sets` - no reason to add code, it's just technical debt for no reason, the speed optimization is not needed here.

PR #246, opened 2019-11-21
SA22-029 Rename also in comments

ogorodnik:
Added:

 - renameInComments configuration option to control whether
    process also comments when renaming or not
 - process comments and includes locations in response when
    comment should be changed
 - new test

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -16,6 +16,7 @@
 ------------------------------------------------------------------------------
 
 with Ada.Characters.Handling; use Ada.Characters.Handling;
+with Ada.Strings.Wide_Wide_Unbounded;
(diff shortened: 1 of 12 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -1521,6 +1522,49 @@ package body LSP.Ada_Handlers is
          Definition : Defining_Name;
          Imprecise  : Boolean;
          Empty      : LSP.Messages.TextEdit_Vector;
+
+         procedure Process_Comments
+           (Node : Ada_Node;
+            Uri  : LSP.Messages.DocumentUri);
+         --  Iterate over all comments and include them in the responce when
+         --  they contain a remaned word
```
setton: *response
*renamed

ogorodnik: done

PR #259, opened 2019-12-09
SC06-075 Separate Documents from Contexts

setton:
Documents should not belong to specific contexts. For instance, a
document might represent an Ada source which is common to several
project contexts. Following this, a document should not own its
Analysis_Unit, the Analysis_Unit is valid only for a specific context.

Refactor the code in light of this: the documents are now held by
the Message_Handler.

Add a test to check that references for an entity that's common to several
project works after editing the document.

### DIFF
--- a/source/ada/lsp-ada_handlers.ads
+++ b/source/ada/lsp-ada_handlers.ads
@@ -19,9 +19,12 @@
 --  language.
 
 with Ada.Containers.Doubly_Linked_Lists;
+with Ada.Containers.Hashed_Maps;
 
 with GNATCOLL.VFS;    use GNATCOLL.VFS;
 with GNATCOLL.Traces;
+with LSP.Ada_Contexts;
+with LSP.Ada_Documents;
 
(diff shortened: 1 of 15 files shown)


source/ada/lsp-ada_handlers.ads
```diff
@@ -117,6 +128,9 @@ private
 
       Refactoring : Refactoring_Options;
       --  Configuration options for refactoring
+
+      Open_Documents : Document_Maps.Map;
```
AnthonyLeonardoGracio: Is it really needed? I don't see any references to this map in this patch

setton: yes, you need to expand the diff to `source/ada/lsp-ada_handlers.adb` - the github interface hides it by default because it's too large (big facepalm, github UI designers 🤦‍♀)

AnthonyLeonardoGracio: Indeed  🤦‍♀

PR #260, opened 2019-12-10
SA22-030: Filter out 'end label' references

AnthonyLeonardoGracio:


source/ada/lsp-ada_handlers.adb
```diff
@@ -1000,6 +1001,12 @@ package body LSP.Ada_Handlers is
         (Node : Ada_Node) return LSP.Messages.AlsReferenceKind_Set;
       --  Fetch reference kind for given node
 
+      function Is_End_Label (Node : Ada_Node) return Boolean
+      is
+         (not Node.Parent.Is_Null and then Node.Parent.Kind in Ada_End_Name);
```
setton: I think the test needs to be improved a little bit for the case the first parent is a DottedName which is itself child of the EndName - for instance in `end <package>.<subpackage>`

AnthonyLeonardoGracio: Ok, I'll check this

PR #262, opened 2019-12-10
SC06-071 Add a valgrind mode to the testsuite

setton:
Add a mode to run the testsuite under valgrind.

To do this, the command line to run the language server
must be modifiable: no longer read it from the .json files,
but read it from the environment directly.

### DIFF
--- a/testsuite/testsuite.py
+++ b/testsuite/testsuite.py
@@ -33,6 +33,10 @@ class ALSTestsuite(Testsuite):
             help="Compute the source code coverage of testcases on ALS. This"
                  " requires GNATcoverage working with instrumentation and will"
                  " run a build of ALS before running tests.")
+        self.main.argument_parser.add_argument(
+            "--valgrind_memcheck", action="store_true",
+            help="Runs the Ada Language Server under valgrind, in memory"
+                 " check mode. This requires valgrind on the PATH.")
 
     def lookup_program(self, *args):
         """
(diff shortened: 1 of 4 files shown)


testsuite/testsuite.py
```diff
@@ -62,8 +66,16 @@ def tear_up(self):
         self.env.repo_base = os.path.abspath(os.path.join(
             os.path.dirname(__file__), '..'))
 
+        self.env.wait_factor = 1
+
         # Absolute paths to programs that test drivers can use
-        self.env.als = self.lookup_program('server', 'ada_language_server')
+        if self.env.options.valgrind_memcheck:
+            self.env.als = "{} --tool=memcheck --quiet {}".format(
```
raph-amiard: Not sure why you want it to be quiet? Having the diff with the locations of the leaks (even passing `--leak-check=full`) would help in nightly runs/CIs IMHO.

pmderodat: Valgrind’s `--quiet` option is used to remove all the messages that are not memory issues diagnostics: leaks, invalid reads and so on (with their stack traces) are still printed.

raph-amiard: @pmderodat aah ok! Still, in this setup you won't get all the diagnostics that you get from `--leak-check=full` right?

pmderodat: You’re right: `--leak-check=full` would help to find more memory leaks.

setton: ah, note that this mode "memcheck" is going after memory corruption, not leaks! I want to only display a message if there's an error. I will add a mode valgrind_leaks for the leaks, in a separate PR.

raph-amiard: Why doing two different modes?  `leak-check=full` is the same checker (memcheck) and will also detect memory corruptions

PR #275, opened 2019-12-16
SA14-007 list parent methods in 'definition'

setton:


doc/reference_kinds.md
```diff
@@ -13,12 +13,14 @@ extra field to the `Location` type:
...
-export type AlsReferenceKind = 'write' | 'call' | 'dispatching call';
+export type AlsReferenceKind = 'write' | 'call' | 'dispatching call' | 'parent' | 'child';
 
 export namespace AlsReferenceKind {
    export const Write            : AlsReferenceKind = 'write';
    export const Static_Call      : AlsReferenceKind = 'call';
    export const Dispatching_Call : AlsReferenceKind = 'dispatching call';
+   export const Parent           : AlsReferenceKind = 'parent';
+   export const Child            : AlsReferenceKind = 'child';
 }
```
AnthonyLeonardoGracio: We use that type to create filters for the 'Find references...' action in GNAT Studio: thus, adding these values here will add two useless/non-active filters to the 'Find references...' dialog.

Two solutions: we can encode this information in a separate type or handle the filters properly in the textDocument/references handler. That can be done in a separate patch though.

setton: I propose to hard-code the values that are of interest for filters in GS.

AnthonyLeonardoGracio: Ok, fine by me

PR #277, opened 2019-12-18
SC05-038 Implement 'textDocument/implementation'

setton:
Generate the boilerplate to support this, and implement the body.

Revisit the way we handle contexts: store the source files in an ordered
set to improve speed of lookup and memory management.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -613,6 +613,90 @@ package body LSP.Ada_Handlers is
       return Response;
    end On_Declaration_Request;
 
+   -------------------------------
+   -- On_Implementation_Request --
+   -------------------------------
+
+   overriding function On_Implementation_Request
+     (Self    : access Message_Handler;
+      Request : LSP.Messages.Server_Requests.Implementation_Request)
+      return LSP.Messages.Server_Responses.Location_Response
+   is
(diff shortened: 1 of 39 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -613,6 +613,90 @@ package body LSP.Ada_Handlers is
...
+         --  Find the definition
+         Definition := Resolve_Name (Name_Node, Self.Trace, This_Imprecise);
+         Imprecise := Imprecise or This_Imprecise;
+
+         --  If we didn't find a definition, give up for this context
+         if Definition = No_Defining_Name then
+            return;
+         end if;
+
+         --  Now that we have a definition, list all the implementations for
```
AnthonyLeonardoGracio: Looking at https://stackoverflow.com/questions/58380425/whats-the-difference-between-textdocument-definition-textdocument-implementati, I think that we should implement the same mechanism as for textDocument/definition here (i.e: list base and overriding subprograms).

setton: Allright! let’s do this as part of a subsequent PR, if OK with you?

AnthonyLeonardoGracio: Yes, np for me

PR #279, opened 2019-12-19
SC19-050 Fix textDocument/typeDefinition

setton:
This was not working in the context of aggregate projects,
and was not working when requested on the definition of the
type itself, only on an use of the type.

Add a test that covers both cases.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -839,40 +839,78 @@ package body LSP.Ada_Handlers is
    is
       use Libadalang.Analysis;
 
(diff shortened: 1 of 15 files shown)


AnthonyLeonardoGracio:
There are 3 commits with differents tickets under the same PR: can we split this PR? It will make the reviews a bit easier IMO.

setton:
> There are 3 commits with different tickets under the same PR: can we split this PR? It will make the reviews a bit easier IMO.

These are the changes from the previous PRs... they are not independent. But you can review only the last commit in each PR

PR #286, opened 2019-12-26
Topic/rename provider

reznikmm:


source/protocol/lsp-generic_optional.adb
```diff
@@ -35,6 +35,17 @@ package body LSP.Generic_Optional is
    begin
       if Value.Is_Empty then
          V := (Is_Set => False);
+      elsif Value.Kind in GNATCOLL.JSON.JSON_Boolean_Type then
+         --  During protocol extension some boolean settings become objects.
+         --  To keep reading of optional settings compatible with earlier
+         --  implementations we read `true` (while we expect an optional
```
setton: I don't understand why we do this - I think we should just support one version of the protocol, not support earlier ones. Or is there a good reason to do this?

reznikmm: `ALS` supports 3.14 while `clangd` still uses older one. GPS should be able to read both.

setton: OK, so I have a question. With this change, we return

    hoverProvider: {}

in the version 3.14 of the protocol, the spec says:

	/**
	 * The server provides hover support.
	 */
	hoverProvider?: boolean;

and in 3.15, it says:

	/**
	 * The server provides hover support.
	 */
	hoverProvider?: boolean | HoverOptions;

So, returning `{}` will break compatibility with clients that implement 3.14, right? Does this break VScode?

ogorodnik: In general it is not related to clangd but to LSP protocol which says following:
	/**
	 * The server provides rename support. RenameOptions may only be
	 * specified if the client states that it supports
	 * `prepareSupport` in its initial `initialize` request.
	 */
	renameProvider?: boolean | RenameOptions;

So, we should support both boolean and RenameOptions in this field, and even "new" servers is allowed to send boolean here.

PR #290, opened 2020-01-02
Update testsuite to the new gnatcov interface

pmderodat:
Hello,

Note that the corresponding change in GNATcoverage is not merged yet, so please let me merge this one so that I can merge both in sync. Thanks in advance!


pmderodat:
Thanks for the review, Anthony! :-)

PR #294, opened 2020-01-10
Simplify coverage handling in the testsuite

pmderodat:


gnat/spawn.gpr
```diff
@@ -20,6 +20,13 @@ project Spawn is
    type OS_API_Kind is ("unix", "osx", "Windows_NT");
    OS_API : OS_API_Kind := external ("OS", "unix");
 
+   --  By default, treat warnings as errors in dev mode, but not in prod
+   --  mode. Let users override this default using the WARN_ERRORS environment
```
AnthonyLeonardoGracio: You meant 'SPAWN_WARN_ERRORS' in the comment, no?

pmderodat: This is a copy of the similar block in `lsp.gpr`. I’ll fix both, thanks!

pmderodat: This is now fixed.

PR #300, opened 2020-01-14
SB25-033 Remove performance bottleneck

setton:
When returning a list of locations, we were checking at each insertion
whether the location was already present in the results. This has
a cost of O(n^2)

Instead, append the locations, and sort/remove duplicates only at
the end, changing the cost to O(n*log(n)).

Sort the results by file then location; when sorting files in the
same directory, use a slightly more logical sort than an alphabetic
sort, returning locations in the .ads before the .adb, and the parent
package before the child packages.

### DIFF
--- a/source/ada/lsp-lal_utils.adb
+++ b/source/ada/lsp-lal_utils.adb
@@ -74,12 +74,145 @@ package body LSP.Lal_Utils is
         LSP.Lal_Utils.Get_Node_Location
           (Libadalang.Analysis.As_Ada_Node (Node), Kind);
    begin
-      if not Is_Synthetic
-        and then not Result.Contains (Location)
-      then
+      if not Is_Synthetic then
          Result.Append (Location);
       end if;
    end Append_Location;
+
+   --------------------------------
+   -- Sort_And_Remove_Duplicates --
+   --------------------------------
+
+   procedure Sort_And_Remove_Duplicates
+     (Result : in out LSP.Messages.Location_Vector)
+   is
+      function URI_Inf (Left, Right : LSP.Types.LSP_String) return Boolean;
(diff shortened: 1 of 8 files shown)


source/ada/lsp-lal_utils.adb
```diff
@@ -74,13 +74,146 @@ package body LSP.Lal_Utils is
...
+               end if;
+
+               if L (L'First + Ind) /= R (R'First + Ind) then
+                  --  Return "pack.adb" before "pack-child.adb"
+                  if L (L'First + Ind) = '-'
+                    and then R (R'First + Ind) = '.'
+                  then
+                     return False;
+                  elsif L (L'First + Ind) = '.'
+                    and then R (R'First + Ind) = '+'
```
reznikmm: '+' -> '-' typo?

setton: yes, good catch! will fix right away

source/ada/lsp-lal_utils.adb
```diff
@@ -74,13 +74,146 @@ package body LSP.Lal_Utils is
...
+         --  We're being a bit clever when comparing two URIs:
+         --    * for a same file, return ".ads" before ".adb"
+         --    * return "pack.adb" before "pack-child.adb"
+
+         if L_Dir'Length /= R_Dir'Length then
+            --  The directories are different: compare them
+            return L_Dir < R_Dir;
+         else
+            --  The directories have the same length: are they the same?
+            for Ind in 1 .. L_Dir'Length - 1 loop
```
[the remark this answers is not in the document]
setton: Right, it's equivalent, but iterates through the strings once rather than twice.

PR #312, opened 2020-01-22
SA07-046 Rewrite "Go to type definition" request

reznikmm:
to work with more declaration kinds (not just with object_declaration).


reznikmm:
Done

PR #315, opened 2020-01-24
T123-048 Implement support for receiving incremental editor changes

setton:
For now the data structure implementing the in-memory contents
of the editor is an Unbounded_Wide_String. To speed up the lookup
of locations given in line,character coordinates, add an index
containing the line numbers.

Add a test for some edge cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ ENV/
 
 # Object directory
 .obj
+
+# Expected testsuite output
+testsuite/.als/inout.txt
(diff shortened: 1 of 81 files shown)


reznikmm:
I've changed my mind, we should'nt support full text update since we implement incremental text updates. So, we need to change tests to send incremental `textDocument/didChange` notification.

setton:
I think it would be a big amount of work to redo all the tests that send the full text... And it doesn't hurt to support both modes, does it?

PR #325, opened 2020-02-03
SB05-051 Consider runtime files as part of project files

setton:
The runtime files should be considered when considering
which contexts are appropriate for a given URI. Do this
by maintain a cache of runtime files.

This fixes requests such as 'called_by' when initiated from
the runtime.

NOTE: it's not possible to add a test in this repository in the
current architecture, since the location of runtime files is a
moving target. I will propose a test as part of GNAT Studio.

### DIFF
--- a/source/ada/lsp-ada_context_sets.ads
+++ b/source/ada/lsp-ada_context_sets.ads
@@ -50,29 +50,28 @@ package LSP.Ada_Context_Sets is
    --  Return the first context in Contexts which contains a project
    --  which knows about file. Fallback on the "no project" context.
 
-   function Contexts_For_URI
-     (Self : Context_Set'Class;
-      URI  : LSP.Messages.DocumentUri) return Context_Lists.List;
-   --  Return a list of contexts that are suitable for the given URI:
-   --  a list of all contexts where the file is known to be part of the
-   --  project tree. If the file is not known to any project, return
-   --  an empty list.
-   --  The result should not be freed.
-
    function Total_Source_Files (Self : Context_Set'Class) return Natural;
    --  Number of files in all contexts
 
    procedure Cleanup (Self : in out Context_Set'Class);
    --  Free memory referenced by Self
 
-   function Each_Context (Self : Context_Set)
-     return Context_Lists.List_Iterator_Interfaces.Forward_Iterator'Class;
-   --  Iterate over contexts of the set
-
    function Get
      (Self : Context_Set;
       Id   : LSP.Types.LSP_String) return Context_Access;
    --  Return context by its Id
+
+   type Context_Predicate is access function
+     (Context : LSP.Ada_Contexts.Context) return Boolean;
+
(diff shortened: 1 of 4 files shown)


source/ada/lsp-ada_context_sets.ads
```diff
@@ -50,15 +50,6 @@ package LSP.Ada_Context_Sets is
    --  Return the first context in Contexts which contains a project
    --  which knows about file. Fallback on the "no project" context.
 
-   function Contexts_For_URI
```
reznikmm: I would propose to keep this function in lsp-ada_context_sets.ads

source/ada/lsp-ada_handlers.adb
```diff
@@ -165,6 +175,54 @@ package body LSP.Ada_Handlers is
...
+
+   function Contexts_For_URI
+     (Self : access Message_Handler;
+      URI  : LSP.Messages.DocumentUri)
+      return LSP.Ada_Context_Sets.Context_Lists.List
+   is
+      function All_Contexts return LSP.Ada_Context_Sets.Context_Lists.List;
+      --  All known contexts
+
+      function All_Contexts return LSP.Ada_Context_Sets.Context_Lists.List is
```
reznikmm: And add All_Contests to lsp-ada_context_sets.ads also.

source/ada/lsp-ada_handlers.adb
```diff
@@ -165,6 +175,54 @@ package body LSP.Ada_Handlers is
...
+
+      File    : constant Virtual_File := To_File (URI);
+      Result  : LSP.Ada_Context_Sets.Context_Lists.List;
+      Context : Context_Access;
+   begin
+      --  If the file does not exist on disk, assume this is a file
+      --  being created and, as a special convenience in this case,
+      --  assume it could belong to any project.
+      if not File.Is_Regular_File then
+         return All_Contexts;
```
reznikmm: Then this will call Self.Contexts.All_Contexts from  lsp-ada_context_sets.ads

source/ada/lsp-ada_handlers.adb
```diff
@@ -165,6 +175,54 @@ package body LSP.Ada_Handlers is
...
+      end if;
+
+      --  If the file is a runtime file for the loaded project environment,
+      --  all projects can see it.
+      if Self.Project_Predefined_Sources.Contains (File) then
+         return All_Contexts;
+      end if;
+
+      for E in Self.Contexts.Each_Context loop
```
reznikmm: Here will became a call to Self.Contexts.Contexts_For_URI from  lsp-ada_context_sets.ads

setton: This doesn't work, unfortunately: the function Contexts_For_URI needs to know the runtime files for the loaded project tree: this is owned by the Ada_Handler, not by the Context_Set.

Issue #309, opened 2020-01-21
Proposal for Collaboration


Entomy:
This is not something I have come to easily. However, it is something I beleive to overall be in the best interest for the community, and each other.

Let's cover hard facts and intended outcome first, and then my intentions after.

At the time this extension was formally announced and uploaded to the marketplace, there were already four Ada extensions in existence, at least that I'm aware of. Two of these were already on the marketplace, the oldest seeming to belong to Alessandro Del Sole, as well as my own. The other two that I'm aware of haven't been uploaded to the marketplace, but belong to Yuta Tomino (sp?) and Luke Guest.

Despite Ada being quite heavily used in certain industries, Ada's FOSS community is small. Extremely small. This kind of fragmentation isn't doing any good, and does more to hurt each other than to drive anything forward.

So now use and growth for the marketplace extensions:

* Del Sole's
    * 3,524 installs
    * Averaging 2.53 installs per day
* Mine (not including separate snippets extension)
    * 7,383 installs
    * Averaging 10.24 installs per day
* Yours
    * 312 installs
    * Averaging 3.00 installs per day

I've overwhelmingly focused on the syntax classifier, as it's closest to the expertise I've spent a decade and a half developing. I haven't had any desire nor time required to implement an LSP. While the initial implementation was created while also learning how VS Code does classification, it came out decently well, and was able to classify more precisely than even GPS, just with certain limitations imposed by the engine the classifier uses. More recently I've been made aware of ways to work with and work around those limitations to get near close to the EBNF definitions, and handle line breaks and other limitations much more robustly. I'll get back to this later.

Yours, on the other hand, does have LSP support. But also far from great classification. 

What I'm proposing is that the two projects get combined. Your LSP, my syntax classifier.

This would have a number of benefits, not including the few related to my intentions, which will be covered later.

1. **Reduced fragmentation of tooling.** As stated, this doesn't do the community any good. It splits resources in an already small community.
2. **Reduced fragmentation of community.** The subset of Ada programmers who are also VS Code users is even smaller. Fragmenting this community causes even more problems. The bizarre thing is, for highly utilized languages like C and C++, there's only one language extension in the marketplace (until you hit the back pages; wouldn't recommend that in general), with a number of related extensions adding in features on top, not recreating the extension.
3. **Your team can focus more on LSP development.** I shouldn't need to explain this, right? It reduces fragmentation of resources within your own team.
[...]

PR #333, opened 2020-02-13
Merging Projects

Entomy:
**Don't merge this yet**

I'm opening the PR now to allow for individual components to be assessed and merged/rejected much easier than if the entire thing was done all at once. There's several conflicting files, such as the ada/gpr definitions, but also several new files, and different naming schemes for things, that will all need to be addressed.

As stated in #309, part of why I'm doing this is that I simply can't continue to use VS Code given how parts of a disability has progressed, and how the VS Code team doesn't seem interested in a solution, which is insane because I know I'm not the only person troubled by use of a mouse.

The files I'm merging are from the last state of the repo that made it up on the marketplace. I've worked out better solutions for the overall approach and organization, and have partially implemented that work. I can continue to try to work on them in the future, but I also have no issue sharing that knowledge if someone else feels like taking a stab at it or helping.

### DIFF
--- a/integration/vscode/ada/ada-snippets.json
+++ b/integration/vscode/ada/ada-snippets.json
@@ -68,7 +68,7 @@
 		"scope": "ada",
 		"body": [
 			"elsif ${1:Condition} then",
-			"\t$0"
+			"   $0"
 		],
 		"description": "Elsif Statement"
     },
@@ -85,9 +85,9 @@
 		"scope": "ada",
 		"body": [
 			"entry ${1:Name} when ${2:Guard_Condition} is",
-			"\t$3",
+			"   $3",
 			"begin",
-			"\t$0",
+			"   $0",
 			"end ${1:Name};"
 		],
 		"description": "Entry Body"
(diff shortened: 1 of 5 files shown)


Entomy:
That one's got what I've worked out of the `.ali` syntax. It's limited, but provides some highlighting.

Entomy:
Well, you guys already have my `.gpr` syntax 😂. Thanks for fixing the inconsistent whitespace from before I added an `.editorconfig`

setton:
@Entomy thank you for this! Waiting for your green light before we review/merge.

(@dgautherot you forgot to credit @Entomy for `gpr.tmLanguage.json` if you had imported it from his repository :-))

Entomy:
@setton legally speaking, he's fine and actually didn't need to in this instance. I released that code in the public domain. He owns it just as much as I do.

@dgautherot you still should as a "cover your ass" decision 😉

Entomy:
@setton 🟢

setton:
Thanks a lot for this! This all looks good to me, but I'm not an expert, I would like @reznikmm to have a look as well, since he knows the VScode integration better.

Entomy:
> Well, you guys already have my .gpr syntax 😂.

Since this is inevitably one of the dangerous sides of OSS development, [Codequiry](https://codequiry.com/) may be useful. It's derived from [MOSS](https://theory.stanford.edu/~aiken/moss/) and checks against some online sources as well. Although it's still primarily meant for academic situations, they do list use cases for outsourcing and code infringement claims.

Entomy:
I'll add some closing remarks so everyone's on the same page.

With regards to language support, a large portion of Ada syntax, definitely most things people use, is supported. This should include your **gnatdoc** comments as well, albeit only simply, and iirc, your convention of comment boxes above subroutines (I used them as section markers, but the syntax is the same). Jere has pointed out a few cases where little to no highlighting occurs. Iirc record layout clauses are one of them. I delayed this for these rewrite.

The syntax files are severely nested. When I was writing the extension initially this was allowed by the schema. This is still legal, but discouraged, now. Unnesting this is probably the best starting point towards refining them.

VS Code's classifier operates on a line by line basis, using an implementation of Oniguruma Regex. As such, allowing line breaks inside of syntax elements becomes interesting. The existing implementation is spotty, but does conform to most common style conventions. Through clever use of `begin` and `end`, along with lookarounds, you can get it to allow line breaks at _almost_ every keyword, identifier, and delimiter boundary. I've done some of the initial work there, to prove it works, at least. 

With regards to theme, as I developed the extension I did test against the built in themes, but also developed my own theme [Ofbase](https://marketplace.visualstudio.com/items?itemName=Entomy.offbase), which supports the more fine grained highlighting that benefits keyword heavy languages like Ada. Default themes work fine regardless, although I do find Microsoft's themes "Light+/Dark+" do make the bad assumption that intrinsic types are keywords, and are themed the same way, so those specific themes can cause entire lines to look like keywords. If you check TM Scope you can see it's a theme oddity, but user assumption may be extension error. Going forward it may be a good idea to slightly violate convention and classify all types as `support.type` regardless, rather than `storage.type`.

setton:
Thank you for the valuable insights! Hope VS Code won’t forbid nesting before we have had time to unnest.

setton:
And thank you for the valuable contributions @Entomy!

PR #335, opened 2020-02-13
Added basic .editorconfig, closes #334

### LINKED ISSUE #334: Add .editorconfig
You guys use unique styling conventions. By adding an [`.editorconfig`](https://editorconfig.org/) file all supporting editors will automatically use the appropriate project style conventions, which reduces contributor and reviewer friction. Supporting editors include [VS Code](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig), [Emacs](https://melpa.org/#/editorconfig), [Vim](https://www.vim.org/scripts/script.php?script_id=3934), and more. There's no negative impact to this. Non supporting editors will adhere to their user preferences, and supporting editors only use the `.editorconfig` preferences in that workspace, preserving user settings otherwise.

Entomy:
I sampled about 14 files in different directories and came up with this for general rules. Seems like you guys have some python specific rules, but I don't know python.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 3
+charset = utf-8
+end_of_line = crlf
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,6 @@ root = true
 indent_style = space
 indent_size = 3
 charset = utf-8
-end_of_line = crlf
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 3
+charset = utf-8
+end_of_line = crlf
```
setton: `end_of_line` should be `lf` for anything edited manually - maybe the counting was thrown off by the fact that the protocol uses `crlf` ?

PR #344, opened 2020-02-24
T213-040 Remove the type Context_Access.

setton:
For memory safety, prefer to no longer make this type limited, so it
can be stored in Ada containers, and remove the use of access types
everywhere.

### DIFF
--- a/source/ada/lsp-ada_context_sets.adb
+++ b/source/ada/lsp-ada_context_sets.adb
@@ -15,8 +15,6 @@
 -- of the license.                                                          --
 ------------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


reznikmm:
Actually latest SPARK supports access types AFAIK.

setton:
> Actually latest SPARK supports access types AFAIK.

Yes it does, with ownership rules. But in this case access types are not necessary, let's remove them wherever we can.

reznikmm:
I don't think these Context_Access is unnecessary. For instance Context contains cache for list of source files. Copying this cache could be expensive and the cache doesn't have sense. We can just query this set from Project_Tree instead.

PR #347, opened 2020-02-24
T224-034 Rework support for canceling requests

setton:
Use a protected object for data that's shared between the Input_Task
and the Processing_Task. Remove the use of a set of pointers in
Request_Map.

### DIFF
--- a/source/server/lsp-servers.ads
+++ b/source/server/lsp-servers.ads
@@ -35,7 +35,7 @@ with LSP.Types;
 with GNATCOLL.Traces;
 
 private with Ada.Strings.Unbounded;
-private with Ada.Containers.Hashed_Maps;
+private with Ada.Containers.Hashed_Sets;
 private with Ada.Containers.Synchronized_Queue_Interfaces;
 private with Ada.Containers.Unbounded_Synchronized_Queues;
 private with GNAT.Semaphores;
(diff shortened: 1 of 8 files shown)


source/server/lsp-servers.ads
```diff
@@ -235,14 +281,15 @@ private
       Processing_Task : Processing_Task_Type (Server'Unchecked_Access);
       Output_Task     : Output_Task_Type (Server'Unchecked_Access);
       Input_Task      : Input_Task_Type (Server'Unchecked_Access);
-      Request_Map     : Request_Maps.Map;
```
reznikmm: Request_Map is used only in Input_Task. There is no concurrency conflict there

setton: Indeed.

setton: To be clear: I still want to remove the pointer in Request_Maps, since I want to remove the type Message_Access - I'm doing this step by step.

source/server/lsp-servers.ads
```diff
@@ -125,6 +125,14 @@ package LSP.Servers is
    --  notification/request it's currently processing. This should only be
    --  called from the processing task.
 
+   --------------------------
+   -- Cancellation support --
+   --------------------------
+
+   function Cancel_Current (Self : Server) return Boolean;
```
reznikmm: I'm afraid making server to have just one _current_ request could be too restrictive for us. In 3.15 we have _partial result_, so server can execute several request at once. Consider long running "find all references" and working tooltips during its execution. It's very attractive to have.

setton: no, it will never be possible to have the *processing* task execute more than one current request: LAL is not task safe

reznikmm: We could do this way

1. take 10 files from Source_Files and execute 'find references' for them
2. send partial results and progress report for these 10 files
3. check if we have 'over' or similar request and execute it if any
4. continue with next 10 files, etc

setton: yes, that could be interesting! will go back to the drawing board.

PR #349, opened 2020-02-25
T210-007 Add a script which digests traces

setton:
For the purpose of debugging and investigating memory issues.

### DIFF
--- a/scripts/digest_traces.py
+++ b/scripts/digest_traces.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+""" This script processes a traces file: it extracts the requests that seem
+    to be consuming memory.
+
+    ALS.IN and ALS.OUT need to be active for this to work.
+
+    Usage:
+
+        digest_traces.py <traces_file>
+"""
+
+import json
+import os
+import re
+import sys
+
+TRACES_LINE = \
+  ' *\\[ALS\\.(IN|OUT)\\] ({.*}) \\[Watermark:. (\\d+)/ (\\d+)\\].*'
+# Regexp to parse one line of traces
+
+
+class Digester(object):
+    """ The class that takes care of the file parsing and counting """
+
+    def __init__(self, filename):
+        """Initialize. filename is the path to the traces file"""
+
+        self.in_re = re.compile(TRACES_LINE)
+
+        self.last_method = "(elaboration)"
+        self.last_memory_usage = 0
+        # the last method and last memory usage received
+
+        self.methods_to_memory = {}
+        # cumulative memory increase indexed by method id
+
+        self.methods_exception_count = {self.last_method: 0}
+        # count the exception occurrences for each exception
+
+        self.currently_looking_at_exception = False
+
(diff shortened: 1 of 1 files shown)


scripts/digest_traces.py
```diff
@@ -0,0 +1,110 @@
...
+            print("{}:{}{}\t{}\t{}".format(j[1],
+                                           " " * max(1, (40 - len(j[1]))),
+                                           self.methods_count[j[1]],
+                                           self.methods_exception_count[j[1]],
+                                           j[0]))
+        print("\nTOTAL:  {}\n".format(total))
+
+
+def main():
+    if len(sys.argv) < 2:
```
AdrienBoulanger: If it only work with one argument as the input trace file then it should be "if len(sys.argv) != 2:"

scripts/digest_traces.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # read the file
+        with open(filename, "r") as f:
+            for line in f.readlines():
+                self.read_line(line.strip())
+
+    def read_line(self, line):
+        """Read and process one line of traces"""
+        if (not self.currently_looking_at_exception and
+           line.startswith("_ALS.MAIN")):
```
AdrienBoulanger: You are assuming that any multi-lines trace in "ALS.MAIN" is an exception.

scripts/digest_traces.py
```diff
@@ -0,0 +1,110 @@
...
+        self.in_re = re.compile(TRACES_LINE)
+
+        self.last_method = "(elaboration)"
+        self.last_memory_usage = 0
+        # the last method and last memory usage received
+
+        self.methods_to_memory = {}
+        # cumulative memory increase indexed by method id
+
+        self.methods_exception_count = {"(elaboration)": 0}
```
AdrienBoulanger: You should replace the hard-coded "(elaboration)" by self.last_method

scripts/digest_traces.py
```diff
@@ -0,0 +1,110 @@
...
+        # count the exception occurrences for each exception
+
+        self.currently_looking_at_exception = False
+
+        self.methods_count = {"(elaboration)": 1}
```
AdrienBoulanger: Same as previous

scripts/digest_traces.py
AdrienBoulanger: This script can only handle one argument => the condition should be len(sys.argv) != 2

scripts/digest_traces.py
setton: yes, in this version we have no better way of doing this! I will modify the ALS to say "EXCEPTION" shortly.

scripts/digest_traces.py
setton: done

scripts/digest_traces.py
setton: done

scripts/digest_traces.py
setton: done

PR #351, opened 2020-02-28
SA22-032 Code folding

ogorodnik:


source/ada/lsp-ada_documents.ads
```diff
@@ -94,6 +95,18 @@ package LSP.Ada_Documents is
       return Libadalang.Analysis.Ada_Node;
    --  Get Libadalang Node for given position in the document.
 
+   function Get_Root_Node
+     (Self    : Document;
+      Context : LSP.Ada_Contexts.Context)
+      return Libadalang.Analysis.Ada_Node;
+   --  Get Libadalang Root node for the given document.
+
+   function Get_First_Token
```
reznikmm: I don't see where `Get_First_Token` is called. Do we really need it?

source/ada/lsp-ada_documents.ads
```diff
@@ -94,6 +95,18 @@ package LSP.Ada_Documents is
       return Libadalang.Analysis.Ada_Node;
    --  Get Libadalang Node for given position in the document.
 
+   function Get_Root_Node
```
reznikmm: May I suggest to not expose the root node, but move folding population code into the `Ada_Documents` instead? It would make Ada_Handler a bit shorter.

source/ada/lsp-ada_handlers.adb
```diff
@@ -560,6 +565,15 @@ package body LSP.Ada_Handlers is
          Self.Get_Symbols := LSP.Ada_Documents.Get_Symbols'Access;
       end if;
 
+      if Value.capabilities.textDocument.foldingRange.Is_Set then
+         if Value.capabilities.textDocument.foldingRange.Value.
```
reznikmm: Use `and then` instead of `then if`?

source/ada/lsp-ada_handlers.adb
```diff
@@ -1438,16 +1452,259 @@ package body LSP.Ada_Handlers is
...
+                 Ada_Aspect_Assoc |
+                 Ada_Aspect_Spec =>
+--        Cat_Aspect
+
+               Store_With;
+               foldingRange.kind :=
+                 (Is_Set => True, Value => LSP.Messages.Region);
+
+               Sloc := Sloc_Range (Data (Node.Token_Start));
+               foldingRange.startLine := LSP_Number (Sloc.Start_Line) - 1;
```
reznikmm: Could you reuse `LSP.Lal_Utils.Get_Node_Location` function? This will simplifies job when we will fix Position.column calculation at some point. The offsets are based on a UTF-16 string representation, so counting it as character index is incorrect.

source/ada/lsp-ada_handlers.adb
```diff
@@ -1438,16 +1452,259 @@ package body LSP.Ada_Handlers is
...
+
    begin
-      Response.error :=
-        (True,
-         (code => LSP.Errors.InternalError,
-          message => +"Not implemented",
-          data => <>));
-      return Response;
+      if Doc /= null then
+--           Doc.Get_Root_Node (C.all).Unit.Print;
```
reznikmm: Drop this comment.

source/ada/lsp-ada_handlers.adb
```diff
@@ -1438,16 +1452,259 @@ package body LSP.Ada_Handlers is
...
+      foldingRange : LSP.Messages.FoldingRange;
+      Response     : LSP.Messages.Server_Responses.FoldingRange_Response
+        (Is_Error => False);
+
+      Have_With : Boolean := False;
+
+      procedure Parse (Node : Libadalang.Analysis.Ada_Node);
+      --  Check the Node and all its children recursively
+
+      procedure Parse (Token : Libadalang.Common.Token_Reference);
```
reznikmm: `s/represen t/represent/` or may be _contain_ is better.

source/ada/lsp-ada_handlers.adb
```diff
@@ -1438,16 +1452,259 @@ package body LSP.Ada_Handlers is
...
+               end if;
+
+               foldingRange.endLine := LSP_Number (Sloc.End_Line) - 1;
+               Have_With := True;
+
+            when others =>
+               Store_With;
+         end case;
+
+         declare
```
reznikmm: A proposal: use standard Libadalang traverse procedure to iterate tree nodes, then do a separate pass over trivias-comments instead of inventing a custom traverse code.

Feel free to ignore.

AnthonyLeonardoGracio:
Don't forget to update the README once we have the request fully implemented

Entomy:
This should probably remove the `foldingStartMarker` and `foldingStopMarker` from [ada.tmLanguage.json](https://github.com/AdaCore/ada_language_server/blob/b392300cfd02f2ecdc971b173516284bc1101423/integration/vscode/ada/syntaxes/ada.tmLanguage.json#L4) as well, so they don't conflict with each other.

PR #357, opened 2020-03-02
SA22-032 Corrections for the comments folding

ogorodnik:


ogorodnik:
> Could you please add a test for this?

Done

testsuite/ada_lsp/SA22-032.folding/test.json
```diff
@@ -0,0 +1,173 @@
+[
+   {
+      "comment": [
+         "test automatically generated"
```
setton: this needs to be adjusted to say what the test does

ogorodnik: Done

PR #362, opened 2020-03-08
Normalized indentation

Entomy:
Normalized indentation since it was getting mixed. Added an `.editorconfig` to the `vscode/ada/` dir with some specifics for that directory.


Entomy:
I fixed the README in this as well, since it was a simple change and didn't warrant its own PR. I didn't author the syntax highlighter, just the classifier. A classifier is part of an analyzer, and is basically a tokenizer, while a syntax highlighter is part of an editor. The two talk to each other, but Microsoft gets credit for the syntax highlighter. I did supply the majority of the snippets though. 😉

PR #364, opened 2020-03-11
T311-014: Filter completion results with typed prefix

AnthonyLeonardoGracio:
This allows to return much less results, increasing the
overall performance for this request.


AnthonyLeonardoGracio:
I prefer to add proper tests a bit later: it's a bit too soon, many things will change, including the order of results etc.

PR #371, opened 2020-03-17
T317-049: Remove duplicate results in Is_Called_By request

AdrienBoulanger:
Duplicates were happening when concatening the results from
different contexts.


AdrienBoulanger:
It was discovered on a complex aggregate project, so I could try to blindly create a simple one to see if the issue happens. But I am not sure it will be possible to have a test about it.

AnthonyLeonardoGracio:
> It was discovered on a complex aggregate project, so I could try to blindly create a simple one to see if the issue happens. But I am not sure it will be possible to have a test about it.

I think that we have already tests with aggregate projects in the ALS testsuite, you can try with them maybe

PR #376, opened 2020-03-19
T317-073 increase/reduce subprogram box on rename

ogorodnik:


source/ada/lsp-ada_handlers.adb
```diff
@@ -2114,26 +2115,147 @@ package body LSP.Ada_Handlers is
...
+            Current : Token_Reference;
+            Diff    : Integer;
+
+            Box_Line : constant Wide_String (1 .. 80) := (others => '-');
+
+            procedure Include_Result (Edit : LSP.Messages.TextEdit);
+            --  Add prepared TextEdit to Result;
+
+            function Process_Box return Boolean;
+            --  Check whether Current is box header/fother and modify it.
```
reznikmm: s/fother/footer/ ?

source/ada/lsp-ada_handlers.adb
```diff
@@ -2114,26 +2115,147 @@ package body LSP.Ada_Handlers is
...
             use Libadalang.Common;
 
-            Token : Token_Reference := First_Token (Node.Unit);
-            Name  : constant Wide_Wide_String :=
+            Token   : Token_Reference := First_Token (Node.Unit);
+            Name    : constant Wide_Wide_String :=
               Ada.Strings.Wide_Wide_Unbounded.To_Wide_Wide_String
                 (Get_Last_Name (Name_Node));
-            Span  : LSP.Messages.Span;
+            C       : LSP.Messages.TextEdit;
```
reznikmm: Can you use more characters in names? I propose to replace `C` with `Text_Edit` for example.

source/ada/lsp-ada_handlers.adb
```diff
@@ -2114,26 +2115,147 @@ package body LSP.Ada_Handlers is
...
+            --  Check whether Current is box header/fother and modify it.
+            --  Return False when the searching cycle should be stopped.
+
+            --------------------
+            -- Include_Result --
+            --------------------
+
+            procedure Include_Result (Edit : LSP.Messages.TextEdit) is
+            begin
+               if not Response.result.changes (Uri).Contains (Edit) then
```
reznikmm: Can we ensure unique responses by replacing algorithm itself and avoid dynamic check?

source/ada/lsp-ada_handlers.adb
```diff
@@ -2114,26 +2115,147 @@ package body LSP.Ada_Handlers is
...
+                  return False;
+               end if;
+
+               case Kind (Data (Current)) is
+                  when Ada_Whitespace =>
+                     return True;
+
+                  when Ada_Comment =>
+                     declare
+                        T : constant Text_Type := Text (Current);
```
reznikmm: `T` is too short

source/ada/lsp-ada_handlers.adb
ogorodnik: I am open to suggestions about how it can be done.

reznikmm: Why not just remove this **`if`**?

Can you provide any example where this condition is actually `True`?

source/ada/lsp-ada_handlers.adb
```diff
@@ -2244,17 +2244,17 @@ package body LSP.Ada_Handlers is
                      end loop;
 
                      --  Include corrected comment itself
-                     Include_Result (C);
+                     Include_Result (Text_Edit);
 
                      Current := Next (Token);
                      loop
-                        --  Looking for the box fother
+                        --  Looking for the box foter
```
reznikmm: s/foter/footer/ ?

source/ada/lsp-ada_handlers.adb
ogorodnik: I can't guarantee that it will be always false and because of the fact that nobody except me will solve potential problems here, I would like to prevent them instead of waiting for customer's complains.

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: Agreed

ogorodnik: @AnthonyLeonardoGracio This is already fixed, that why you can see
source/ada/lsp-ada_handlers.adb Outdated

the keyword is Outdated

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: Checking for duplicates can cost a bit if we have too many edits. Moreover, duplicates should not happen, so I am in favor of removing this 'if' since it might hide a bad behavior before this check

ogorodnik: Done

PR #377, opened 2020-03-20
Final rework

Entomy:
I think since this is a draft it shouldn't be mergable until I'm done. I'm creating the PR because a lot is involved in this, and so people can see things are being worked on. In case anyone wants to contribute to the classifier, I would hold off because conflicts with this are gonna be a nightmare to resolve.

I've tried addressing any of the older components individually, but I'm at the point where nothing else can without just gutting all the older approaches. The reason for this is that the older approach was the more "shotgun" style of everything at the top level, unstructured, without context, that nearly every editor classifier uses. I tried adding context to that and it worked well up to a point.

This will finalize flattening everything, using the newer, heavily structured, approach that I've been taking. A great example of how much this improves things is this:

image

Certainly not what you'd expect from a line oriented parser and regex. That's line breaks allowed at almost every single boundary.

I've got to do this for the entire `type` system, as well as providing something to replace `arguments`, `assignment`, `discriminant`, `exception`, `property`, `return`, and `subroutine`. So there's a lot to this.


setton:
Hello @Entomy, just to let you know we're waiting for your green light to review this.

Entomy:
Yeah it's almost done. Just discriminants and record fields left, I think.

We're trying to orchestrate me returning back to medical work to help out with everything going on, so I might not be that available for a while. Not sure yet.

Entomy:
Everything looks fine on my end. However I don't have many Ada sources available to test with. The new approach uses a much more structured approach, not a "shotgun" approach. So it knows about context and can highlight appropriately for that, but also needs to be told about almost every single grammar rule. So I'd check for serious regressions, because there's a good possibility my test sources don't cover some cases.

setton:
Perhaps it's time for us to look into setting up VS Code automated testing - the code is becoming too important to be tested manually. Don't hesitate to send pointers if you happen to have experience with this.

Entomy:
Not with UI. I do entirely backend stuff which is easily tested with standard unit tests.

Entomy:
I did find this (https://developers.redhat.com/blog/2019/11/18/new-tools-for-automating-end-to-end-tests-for-vs-code-extensions/) which seems like what you're trying to do; at least at a basic level.

setton:
This does look promising! mentioned it in the issue. Thank you for opening it!

PR #382, opened 2020-03-23
Topic/sa21 029.document symbols.outline

AdrienBoulanger:
Various enhancements/fixes to adapt the Document_Symbols request for the GS outline

### DIFF
--- a/source/ada/lsp-ada_documents.adb
+++ b/source/ada/lsp-ada_documents.adb
@@ -45,6 +45,13 @@ package body LSP.Ada_Documents is
      (Node : Libadalang.Analysis.Basic_Decl) return LSP.Types.LSP_String;
    --  Return the profile of Node.
 
+   function Is_Declaration
+     (Node : Libadalang.Analysis.Basic_Decl) return Boolean;
+
+   function Get_Visibility
+     (Node : Libadalang.Analysis.Basic_Decl)
+      return LSP.Messages.Als_Visibility;
+
    function Compute_Completion_Detail
      (BD : Libadalang.Analysis.Basic_Decl) return LSP.Types.LSP_String;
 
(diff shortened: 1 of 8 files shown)


source/ada/lsp-ada_documents.adb
```diff
@@ -677,15 +708,141 @@ package body LSP.Ada_Documents is
...
+
+      function To_Profile
+        (Node : Libadalang.Analysis.Subp_Spec'Class)
+         return LSP.Types.LSP_String;
+
+      -------------
+      -- To_Text --
+      -------------
+
+      function To_Text (Node : Ada_Node'Class) return Wide_Wide_String
```
AnthonyLeonardoGracio: Why not return LSP_String directly?

source/ada/lsp-ada_documents.adb
```diff
@@ -677,15 +708,141 @@ package body LSP.Ada_Documents is
...
+                  Result (Cur) := Node_Text (I);
+                  Cur := Cur + 1;
+               end if;
+               Was_Space := True;
+            elsif Node_Text (I) /= ASCII.LF then
+               Was_Space := False;
+               Result (Cur) := Node_Text (I);
+               Cur := Cur + 1;
+            end if;
+         end loop;
```
AnthonyLeonardoGracio: Need comments

source/ada/lsp-ada_documents.adb
```diff
@@ -677,15 +708,141 @@ package body LSP.Ada_Documents is
...
+               end case;
+
+               Append (Item, To_Text (Param.F_Type_Expr));
+
+               if not Init.Is_Null then
+                  Append (Item, " := ");
+                  Append (Item, To_Text (Init));
+               end if;
+
+               for J in Names.First_Child_Index .. Names.Last_Child_Index loop
```
AnthonyLeonardoGracio: Is it really needed? Why can't we just display the subprogram declaration like for tooltips?

source/ada/lsp-ada_documents.adb
AdrienBoulanger: It's easier to concatenate Wide_Wide_String than LSP_String

source/ada/lsp-ada_documents.adb
AdrienBoulanger: The tooltips displaid a multi-lines profile while we want a pretty-printed single line profile.
Let's take the following declaration:

procedure Foo
   (I     : in out Integer;
    Bar : String);

The tooltip will show the declaration itself, while the pretty-printed single line profile is:

Foo (I : in out Integer; Bar : in String)

AnthonyLeonardoGracio: Ok, fine by me then

source/ada/lsp-ada_documents.adb
```diff
@@ -677,15 +721,199 @@ package body LSP.Ada_Documents is
...
+         Cur       : Integer := Node_Text'First;
+      begin
+         for I in Node_Text'Range loop
+            if Node_Text (I) = ' ' then
+               --  Trim multiple whitespace to only keep one
+               if not Was_Space then
+                  Result (Cur) := Node_Text (I);
+                  Cur := Cur + 1;
+               end if;
+               Was_Space := True;
```
AnthonyLeonardoGracio: Use Ada.Strings.Fixed.Trim

AdrienBoulanger: Ada.Strings.Fixed.Trim only handles whitespaces at the beginning and at the end. This is not enough for complex type given in profiles.

In the example below, Trim will not be able to handle properly the type of Bar.
   procedure Foo
       (Bar : not null access function
           (I : Integer;
           J : Float)
           return Boolean);

AdrienBoulanger: Using GNATCOLL.Utils.Split would require to parse each string 2 times:
- First remove the ASCII.LF
- Second Split via whitespace (' ')
(And then we need to concatenate the list from Split)
So regarding performance it would be slower (Note: we can retrieve hundreds of profile)

AnthonyLeonardoGracio: Ok, fine by me then

PR #383, opened 2020-03-24
T306-021: Add memory usage monitor based on GNATCOLL.Memory

AnthonyLeonardoGracio:
when the DEBUG.ADA_MEMORY trace is enabled, we now monitor the
memory usage consumption and dump it in the logs.

This allows to identify which parts of the program are the
biggest memory usage consumers.

The stacktrace addresses are resolved using GNAT.Symbolic.Traceback
because we link the ALS with relocatable libraries.


AnthonyLeonardoGracio:
> See the Travis error :)
> 
> Could you also add a note describing this mechanism in `doc/HACKING.md`?

Done.

AnthonyLeonardoGracio:
I fixed the 'ran' typo too

PR #390, opened 2020-04-05
Add suppression entry for objects allocated during LAL elaboration

setton:


testsuite/leaks.supp
```diff
@@ -60,3 +60,15 @@
    fun:gpr__nmsc__process_naming_scheme__check
 }
 
+# Objects allocated during the elaboration of Libadalang
```
pmderodat, line 63: This comment is misleading. I think we should have instead something such as:
> Valgrind limitation: GNAT’s superaligned secondary stack chunk appears to be leaked, but we actually have a pointer to the middle of the allocated memory block.

PR #391, opened 2020-04-06
T401-049: Fix named parameters command for dot calls

AnthonyLeonardoGracio:
An automatic test has been added too.

### DIFF
--- a/source/ada/lsp-ada_handlers-named_parameters_commands.adb
+++ b/source/ada/lsp-ada_handlers-named_parameters_commands.adb
@@ -176,9 +176,19 @@ package body LSP.Ada_Handlers.Named_Parameters_Commands is
       procedure Append (Params : Libadalang.Analysis.Param_Spec_Array);
       --  Append identifiers from Params to Result
 
+      function Get_Params_Spec_Array
+        (Decl : Libadalang.Analysis.Basic_Decl)
+         return Libadalang.Analysis.Param_Spec_Array;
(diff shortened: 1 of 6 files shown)


source/ada/lsp-ada_handlers-named_parameters_commands.adb
```diff
@@ -217,23 +222,51 @@ package body LSP.Ada_Handlers.Named_Parameters_Commands is
...
-               Append (Params);
+               return Params;
             end;
 
          when Libadalang.Common.Ada_Basic_Subp_Decl =>
             declare
-               Spec : constant Libadalang.Analysis.Base_Subp_Spec :=
+               Spec   : constant Libadalang.Analysis.Base_Subp_Spec :=
                  Decl.As_Basic_Subp_Decl.P_Subp_Decl_Spec;
                Params : constant Libadalang.Analysis.Param_Spec_Array :=
```
AdrienBoulanger, line 232: The Params variables are only used once and can be removed

AnthonyLeonardoGracio: There is no rule that says "You should only create variables when it's used multiple times": variables improve readability and allow easier further refactorings, so we should keep it IMO.

AdrienBoulanger: This is not necessary, just my preference, this is why the PR was also approved

PR #393, opened 2020-04-07
20.2

godunko:


brobecke:
I have restored the branch, based on Vadim telling me that the deletion was not intentional.
I noticed this because I was trying to create the 20-sustained branch and it couldn't find the 20.2 branch.

PR #402, opened 2020-04-15
R519-005: Add a completion test for complex renames

AnthonyLeonardoGracio:
This was not working with the old GNAT Studio completion engine
but now works with the ALS.

### DIFF
--- a/testsuite/ada_lsp/completion.complex_renames/test.json
+++ b/testsuite/ada_lsp/completion.complex_renames/test.json
@@ -1,0 +1,246 @@
+[
+   {
+      "comment": [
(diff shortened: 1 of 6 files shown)


testsuite/ada_lsp/completion.complex_renames/test.json
```diff
@@ -0,0 +1,246 @@
+[
+   {
+      "comment": [
+         "test automatically generated"
```
reznikmm, line 4: Please, provide some test description here

AnthonyLeonardoGracio: Ok

PR #405, opened 2020-04-16
T319-027: Handle package prefixes for subp calls completion

AnthonyLeonardoGracio:
We now use the new libadalang Is_Dot_Call function on the returned
compeltion items to know if we are dealing with a dotted subporgram call
or not, allowing us to distiguish dotted calls, for which we should omit
the first parameter in the returned completion snippets, from subprogram
calls that are prefixed by a package names (e.g: Ada.Text_IO.Put_Line).

An automatic test has also been added.

### DIFF
--- a/testsuite/ada_lsp/completion.subp_params_dotted_names/test.json
+++ b/testsuite/ada_lsp/completion.subp_params_dotted_names/test.json
@@ -1,0 +1,256 @@
+[
+   {
+      "comment": [
+          "This test checks that completion snippets work fine on ",
+          "subprogram calls that are prefixed by a package name."
+      ]
+   },
+   {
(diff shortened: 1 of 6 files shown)


testsuite/ada_lsp/completion.subp_params_dotted_names/test.json
```diff
@@ -0,0 +1,269 @@
...
+               }
+            },
+            "jsonrpc": "2.0",
+            "method": "textDocument/didOpen"
+         },
+         "wait": []
+      }
+   },
+   {
+      "send": {
```
reznikmm, line 159: I would drop this, because it isn't relevant to the test scenario.

PR #406, opened 2020-04-16
Ada syntax

gcluzel:
I'm using vscode with Ada. I made improvements for my own comfort. Maybe they can be useful for everyone. Among them:

- Very partial support for enumeration_representation_clause;
- Rework of type declaration to be able to write `is record` on two different lines;
- Partial support of subtype.

### DIFF
--- a/integration/vscode/ada/syntaxes/ada.tmLanguage.json
+++ b/integration/vscode/ada/syntaxes/ada.tmLanguage.json
@@ -42,12 +42,11 @@
 		},
 		"access_type_definition": {
 			"name": "meta.declaration.type.definition.access.ada",
-			"begin": "(?i)\\b(is)\\s+(not\\s+null\\s+)?(access)\\b",
+			"begin": "(?i)\\b(not\\s+null\\s+)?(access)\\b",
 			"end": "(?i)(?=(with|;))",
 			"beginCaptures": {
-				"1": { "name": "keyword.ada" },
-				"2": { "name": "storage.visibility.ada" },
-				"3": { "name": "storage.visibility.ada" }
+				"1": { "name": "storage.visibility.ada" },
+				"2": { "name": "storage.visibility.ada" }
 			},
 			"patterns": [
 				{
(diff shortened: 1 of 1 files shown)


setton:
Thanks a lot @gcluzel! We'll take a look.

@Entomy @reznikmm is one of you interested in reviewing? Otherwise I'm happy to have a look.

Entomy:
There's a regression as seen here:

image

Whereas the last master:

image

The regression specifically happened with the aspect clause of a package:

image

Which as you'll notice also then introduced a regression with `incomplete_type_definition`.

It otherwise looks good though.

gcluzel:
Thank you @Entomy for your review. I have corrected the regressions you pointed out.

Entomy:
I'm in the ER working. @reznikmm if you can check this, cool, otherwise I'll get to it when I'm out.

Entomy:
Annotation 2020-04-18 081705

Now these regress. Hmm.

You are definately correct in how you split `is record` however.
image 
image

Entomy:
@gcluzel if you can't find an easy solution to this, it might be better to try this as a two stage PR, where `enumeration_representation_clause` and `subtype` are in one, and 
5594a31 is in another. I suspect there's some interplay that's easier to address when isolated.

I got some time before my next shift so I'll try to go through things to help you. I want this to be approved.

gcluzel:
@Entomy I'll split them into three commits and submit three PR which would be easier to understand, you're right.

You noticed my motivation for reworking how types definition are parsed. I'd like to be able to break lines where I want, and not only after a group. `\s` is supposed to match any space characters, including `\n`. But this is not working with my version of vscode and that is really anoying...

Entomy:
@gcluzel that's because of how VS Code works. Parsing is strictly line oriented, and so everything I've done has been based around that. That means `\s` actually doesn't include new lines like you'd expect. So sometimes line breaks aren't allowable between certain keywords in order to actually identify constructs. It's a limitation of the engine.

That being said, they did it for performance reasons, and I haven't ever seen a highlighter get carried away with system resources, so, they did accomplish the goal they set out to solve.

Entomy:
@gcluzel on a related note, I'm glad you figured out how the classifier approach worked. I was worried about that since I can't really document JSON, and it's not exactly a simple approach by any means.

gcluzel:
I took time to understand what was already done. But it is rather clear and well constructed following the Ada RM names!
I'm working on the subtype part. I'll submit another PR. I think that we can close this one.

PR #411, opened 2020-04-20
Add aspect_clause support

gcluzel:
I have added a partial support for the syntax highlighting of `aspect_clause`s.

**Limitations**:

- The support is not complete. The `at_clause` construction is not supported by this work (cf. [here](https://www.adaic.org/resources/add_content/standards/05rm/html/RM-13-1.html#I4417)) because I'm not enough familiar with Ada to know how it is used.

- Moreover, in the following construction, I don't know what scope I should give to `attribute_designator` in `for local_name'attribute_designator use expression;`. Then no scope is assigned to it.

### DIFF
--- a/integration/vscode/ada/syntaxes/ada.tmLanguage.json
+++ b/integration/vscode/ada/syntaxes/ada.tmLanguage.json
@@ -183,6 +183,12 @@
 				"7": { "patterns": [ { "include": "#exponent_part" } ] }
 			}
 		},
+		"basic_declarative_item": {
+			"patterns": [
+				{ "include": "#basic_declaration" },
+				{ "include": "#aspect_clause" }
+			]
+		},
 		"basic_declaration": {
 			"patterns": [
 				{ "include": "#type_declaration" },
@@ -211,7 +217,7 @@
 						"0": { "name": "keyword.ada" }
 					},
 					"patterns": [
-						{ "include": "#basic_declaration" }
+						{ "include": "#basic_declarative_item" }
 					]
 				},
 				{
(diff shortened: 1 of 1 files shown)


gcluzel:
I'd like to have your opinion @Entomy about the way I parse aspect_clause and component_clause in commit ce4562a. I think it's better because lines can be broken everywhere, but I'd like to know what you think about that.

Entomy:
`enumeration_representation_clause` looks good.

`component_clause` should be using:

~~~~ javascript
"beginCaptures": {
    "2": { "name": "storage.modifier.ada" },
},
~~~~

In this context, `at` is specifying a field offset, which modifies the type, so that's more precise about the meaning.

As far as the changes to `aspect_clause` go:

~~~~ javascript
{
    "begin": "(?i)(?<=for)",
    "end": "(?i)(?=use)",
    "captures": {
        "0": { "name": "keyword.ada" }
    },
    "patterns": [
        {
            "match": "((?:\\w|\\d|_)+)(?:(\\')((?:\\w|\\d|_)+))?",
            "captures": {
                "1": { "patterns": [ { "include": "#subtype_mark" } ] },
                "2": { "name": "punctuation.ada" },
                "3": { "name": "entity.other.attribute-name.ada" } // Add this
            }
        }
    ]
}
~~~~

> The support is not complete. The at_clause construction is not supported by this work (cf. here) because I'm not enough familiar with Ada to know how it is used.

Improvements don't need to get every construct flawless from the first commit. We'll get there.

gcluzel:
Thank you @Entomy  for these improvements. I have corrected them

Entomy:
Awesome, thanks!

I can't check this until tomorrow.

Entomy:
I don't have the test sources I've been using here, but I put something together. Looks good to me!

Issue #412, opened 2020-04-21
Highlighting of nested declarations is failing


raph-amiard:
See attached screenshot

![failure](https://user-images.githubusercontent.com/478727/79872727-706b0500-83e6-11ea-9d6e-49a9fd69e0e8.png)

with this following snippet of code

```ada
with Ada.Text_IO;

procedure Main is
   package B is
      procedure C is
      begin
      end C;
   end B;

   procedure D is
   begin
   end D;
begin
end Main;
```

AnthonyLeonardoGracio:
@Entomy I suppose that it's related to your recent syntax highlighting patches no?

Entomy:
Yeah it would be. I can have a look at it tomorrow, as I'm busy for the rest of today.

Entomy:
I checked the file over. `basic_declaration` should include `package_specification` before `subprogram_specification`.

If someone wants to do it before I can, that should do it.

gcluzel:
Ok, I've corrected that as I'm doing some patches. Can someone tell me if a procedure body is allowed in a package spec? GCC rejects this program, and the Ada RM too.
```Ada
with Ada.Text_IO;

procedure Main is
   package B is
      procedure C is
      begin         -- here
      end C;        -- and here
   end B;

   procedure D is
   begin
   end D;
begin
end Main;
```

raph-amiard:
> Ok, I've corrected that as I'm doing some patches. Can someone tell me if a procedure body is allowed in a package spec? GCC rejects this program, and the Ada RM too.

It's illegal indeed. The correct form would be:

```ada
with Ada.Text_IO;

procedure Main is
   package B is
      procedure C;
   end B;

   package body B is
      procedure C is
      begin
         null;
      end C;
   end B;
```

However, and I cannot emphasize that enough: *We don't want to highlight only correct programs*. In light of that, having such a rigid set of rules doesn't seem like a good idea, unless you can fallback on simple keyword highlighting in everycase.

raph-amiard:
FWIW it also wildly fails on highlighting string literals:

![failure2](https://user-images.githubusercontent.com/478727/79959791-d4dca180-8484-11ea-9719-6ff9862a15bf.png)

gcluzel:
What I have done so far is adding procedure calls and resolving bugs that I found in highlighting package body and so on.
![image](https://user-images.githubusercontent.com/34268335/79963845-318e8b00-848a-11ea-9589-a83486ebf4b6.png)
[...]

PR #414, opened 2020-04-22
Improve support for syntax highlighting

gcluzel:
This PR adds features requested in #412. It may introduce conflicts with PR #411.

* Resolve bugs in syntax highlighting
* Adding procedure calls

### DIFF
--- a/integration/vscode/ada/syntaxes/ada.tmLanguage.json
+++ b/integration/vscode/ada/syntaxes/ada.tmLanguage.json
@@ -61,6 +61,20 @@
 				{ "include": "#subtype_mark" }
 			]
 		},
+		"actual_parameter_part": {
+			"begin": "\\(",
+			"end": "\\)",
+			"captures": {
+				"0": { "name": "punctuation.ada" }
+			},
+			"patterns": [
+				{
+					"name": "punctuation.ada",
+					"match": ","
+				},
+				{ "include": "#parameter_association" }
+			]
+		},
 		"array_dimensions": {
 			"name": "meta.declaration.type.definition.array.dimensions.ada",
 			"begin": "\\(",
@@ -114,7 +128,7 @@
 		"aspect_definition": {
 			"name": "meta.aspect.definition.ada",
 			"begin": "=>",
-			"end": "(?i)\\b(?=(,|;|is))\\b",
+			"end": "(?i)(?=(,|;|\\bis\\b))",
 			"beginCaptures": {
 				"0": { "name": "keyword.other.ada" }
 			},
@@ -134,7 +148,7 @@
 		"aspect_specification": {
 			"name": "meta.aspect.specification.ada",
 			"begin": "(?i)\\bwith\\b",
-			"end": "(?i)\\b(?=(;|is))\\b",
+			"end": "(?i)(?=(;|\\bis\\b))",
 			"beginCaptures": {
 				"0": { "name": "keyword.ada" }
 			},
@@ -195,7 +209,8 @@
 				{ "include": "#subtype_declaration" },
 				{ "include": "#exception_declaration" },
 				{ "include": "#object_declaration" },
-				{ "include": "#subprogram_specification" }
+				{ "include": "#subprogram_specification" },
+				{ "include": "#package_declaration" }
 			]
 		},
 		"block_statement": {
@@ -379,6 +394,12 @@
 				"3": { "name": "punctuation.ada" },
 				"4": { "patterns": [ { "include": "#exponent_part" } ] }
 			}
+		},
+		"declarative_item": {
+			"patterns": [
+				{ "include": "#body" },
+				{ "include": "#basic_declarative_item" }
+			]
 		},
 		"delay_statement": {
 			"patterns": [
@@ -677,11 +698,12 @@
 			]
 		},
 		"function_body": {
-			"begin": "(?i)\\b(function)\\s+((\\w|\\d|\\.|_)+)\\b",
-			"end": "(?i)\\b(end)\\s+(\\2)\\s*(;)",
+			"begin": "(?i)\\b(overriding\\s+)?(function)\\s+((\\w|\\d|\\.|_)+)\\b",
+			"end": "(?i)\\b(?:(end)\\s+(\\3)\\s*)?(;)",
 			"beginCaptures": {
-				"1": { "name": "keyword.ada" },
-				"2": { "name": "entity.name.function.ada" }
+				"1": { "name": "storage.visibility.ada" },
+				"2": { "name": "keyword.ada" },
+				"3": { "name": "entity.name.function.ada" }
 			},
 			"endCaptures": {
 				"1": { "name": "keyword.ada" },
@@ -696,7 +718,7 @@
 						"0": { "name": "keyword.ada" }
 					},
 					"patterns": [
-						{ "include": "#basic_declaration" }
+						{ "include": "#declarative_item" }
 					]
 				},
 				{
@@ -707,6 +729,7 @@
 						{ "include": "#statement" }
 					]
 				},
+				{ "include": "#aspect_specification" },
 				{ "include": "#result_profile" },
 				{ "include": "#parameter_profile" }
 			]
@@ -784,7 +807,7 @@
 				},
 				{
 					"begin": "(?i)\\bthen\\b",
-					"end": "(?i)(?=(elsif|else))",
+					"end": "(?i)(?=(elsif|else|end))",
 					"beginCaptures": {
 						"0": { "name": "keyword.control.ada" }
 					},
@@ -929,7 +952,7 @@
 		"package_body": {
 			"name": "meta.declaration.package.body.ada",
 			"begin": "(?i)\\b(package)\\s+(body)\\s+((?:\\w|\\d|\\.|_)+)\\b",
-			"end": "(?i)\\b(end)\\s(\\2)(;)",
+			"end": "(?i)\\b(end)\\s+(\\3)\\s*(;)",
 			"beginCaptures": {
 				"1": { "name": "keyword.ada" },
 				"2": { "name": "keyword.ada" },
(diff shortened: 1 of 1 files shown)


Entomy:
Since you're concerned about possible conflicts, do you want to wait on reviewing this until #411 is committed and you can rebase?

gcluzel:
You can review it without problem. I added the `basic_declarative_item` construction in my both PR because it's more clear when I have it. I can rebase and amend my commit after if needed

integration/vscode/ada/syntaxes/ada.tmLanguage.json
```diff
@@ -183,13 +183,26 @@
 				"7": { "patterns": [ { "include": "#exponent_part" } ] }
 			}
 		},
+		"declarative_item": {
+			"patterns": [
+				{ "include": "#body" },
+				{ "include": "#basic_declarative_item" }
+			]
+		},
```
gcluzel, line 191: This part doesn't work. Whatever the way we put it (`body` first or `basic_declarative_item` first), either `function_body` or `function_specification` is matched.
Maybe the solution is to merge the two definitions. @Entomy you may have better solution.

Entomy: I just got back to where I have my own computer. I'll look at this in about an hour.

Entomy:
@gcluzel will you rebase this? There's a regression and I'm not entirely sure why. That's the easiest first thing to check.

gcluzel:
Sorry, I forgot to push

Entomy:
That was it. Looks good to me!

gcluzel:
Good! But can you just check this snippet
```Ada
with Ada.Text_IO;

procedure Main is
   package B is
      procedure C;
   end B;

   package body B is
      procedure C is
      begin
         null;
      end C;
   end B;

   procedure D
   with
      Import => True;
   
   procedure E is
   begin
      Ada.Text_IO.Put_Line("Hey");
   end E;
begin
end Main;
```
It should fail

Entomy:
🤦🏻‍♂️ oh yeah. I forgot this was related to #412. 

image

This is what I'm seeing. `procedure E` makes sense to fail, because that's not a valid construct, and will have to be handled later on by the failover classifier. `procedure D` on the other hand, is incorrectly classified as a procedure body (`meta.declaration.procedure.body.ada`). Since it breaks beyond this point, this misclassification is almost certainly what's responsible, as that pattern will be looking for an `end D;` which doesn't exist. This is happening because the order of elements within a `patterns` array is important, and you've put matching a procedure body in front of matching a procedure specification. Because the spec hasn't been consumed, and a spec will match the `begin` part of a body, it thinks there's a procedure body there, and handles everything as if there is.

gcluzel:
But `procedure E` is a valid construction in Ada and we have to handle it. In the [Ada RM](https://www.adaic.org/resources/add_content/standards/05rm/html/RM-6-3.html), this construction is allowed and not rejected by gcc.
Then both `procedure D` and `procedure E` have to be handled.

> This is happening because the order of elements within a patterns array is important, and you've put matching a procedure body in front of matching a procedure specification. Because the spec hasn't been consumed, and a spec will match the begin part of a body, it thinks there's a procedure body there, and handles everything as if there is.

I agree with this. This is why I put the `body` scope first, otherwise `package body` is handled as a package declaration.
But this doesn't work for procedures because both constructions exist and both constructions start with the same pattern (which is `(?i)\\b(procedure)\\s+((?:\\w|\\d|\\.|_)+)\\b`). We cannot distinguish which one is supposed to be matched as you mentioned (and even if we put `basic_declarative_item` in first our problem is not solved). So maybe it's necessary to merge both pattern.

Entomy:
> But procedure E is a...

Yeah I was thinking about in a different context. I need to finish my coffee before doing these reviews.

I'll have to think this over, and play with things. I'll get back to you soon on this.

gcluzel:
> I'll get back to you soon on this.

Thank you for your help! And do not hesitate to add your own commits.

Entomy:
Here's an idea:
Put the specs before the bodies, use `is` as the discriminator. That would mean `is` has to be in the `begin` part, but it removes any ambiguity. And all of ARM, GNAT, and my own style have `is` on the same line, so this shouldn't be an issue in practice.

Entomy:
I said that backwards. Using `is` as the discriminator for subroutine bodies requires putting the bodies _before_ the specs, otherwise they'll all be classified as specs.

gcluzel:
I have some reservations about doing that. We also have to match the parameters of the procedure/function. Then a lot of programs won't be match, for example:
```Ada
procedure Add_Descriptor
   (Process : Process_Access;
   Kind    : Pipe_Kinds;
   fd      : Interfaces.C.int)
is
```
But maybe I have another idea. Let's play!

Entomy:
> Then a lot of programs won't be match...

No, you're right, that's an excellent point. Hmm.

> But maybe I have another idea. Let's play!

Oh?

gcluzel:
> Oh?

Finally no. That was a bad idea. I don't see anything else than matching for `procedure_body`
```JSON
"procedure_body": {
	"name": "meta.declaration.procedure.body.ada",
	"begin": "(?i)\\b(overriding\\s+)?(procedure)\\s+((?:\\w|\\d|\\.|_)+)\\b",
	"end": "(?i)(?:\\b(end)\\s+(\\2)\\s*)?(;)",
	....
```
That is very imprecise because we also match incorrect programs. But at least we highlight correctly all correct programs.

setton:
> That is very imprecise because we also match incorrect programs. But at least we highlight correctly all correct programs.

Note that in general it's not a problem to match incorrect programs... as @raph-amiard pointed out, it's something we want. When you are typing things in a source editor, 99% of the time the code is invalid Ada, and we still want the highlighter to do a "good enough guess" highlighting of what's there...

Entomy:
And that's the intention anyways. Much like I said in the issue he raised, there is a fallback mechanism I've always accounted for, right down to exactly how to add it, as well as an explanation of why it's not yet there.

The specific and highly structured nature of the approach taken is to handle context within the language, for an overall more detailed highlighting experience. It's a lot of work just to get some distinction between certain keywords, or actually being able to highlight types in the majority of cases. That being said, like I mentioned there as well, the classifier is not so specific and accurate as to attempt to be a "conforming processor" and certain things have been made more generalized in an effort to keep the classifier working even in failure cases. That may be an option here, where is-begin-end is handled as part of a subroutine declaration internally, rather than a distinct declaration.

The most ideal approach is something I'm not sure how to implement myself, and I'm not sure if you guys are ready to yet. There is a mechanism for an LSP to actually provide classifications of tokens. This is utilized by the TreeSitter based extensions and is the most ideal way to accomplish many of these things, with the tmLanguage files providing quick and easy stuff. To see why, consider the following:

~~~~ ada
declare
    A : Integer := Integer(42);
begin
    -- whatever
end;
~~~~

There is no way, based solely on grammar, to correctly determine `Integer(42)` is a cast and could be classified as a type, and not a subroutine call.

setton:
Given the sophistication of the current engine, I think the first order of priority is to find a way to automate testing for it before we add more sophistication... I'm trying to find the time to have a go at it myself.

gcluzel:
I'm pushing my solution in a separate commit. It's maybe not the best solution, but I don't see any other until a semantic highlighter is implemented. Feel free to give me your feeling about it.

Entomy:
@setton absolutely. The current approach is still correct in the overwhelming majority of cases, and just needs to be developed to completeness. Jumping straight to another design so quickly is foolish.

@gcluzel I won't be able to look until about 21:00 EST, but I will then.

setton:
On my side I started a scaffolding for setting up automated tests. Now I need to learn the API to add real tests in it...

Entomy:
@setton I saw. I'll try to familiarize myself as well and help you out there. With everything else going on I'm not so sure how much time I can devote to this, but I can chip in a bit.

@gcluzel I'm now at the hospital until 24:00 not 21:00, and I've been up since 6:00 so I'll review tomorrow.

gcluzel:
@Entomy Don't worry, there is nothing urgent at all. Take care of you

Entomy:
Oh I'm fine. Just working. Taking care of others.

Entomy:
Looks good, actually. What you did seems to have worked.

On a maybe related note, `goto` and `exit` regressed at some point but I don't know if that happened with this one or elsewhere. Also, I've been noticing the `repository` increasingly getting out of alphabetical order; please keep it that way as it makes scanning through things easier (we can consider another organizational system, that's just what I went with for now.

gcluzel:
> Also, I've been noticing the repository increasingly getting out of alphabetical order; please keep it that way as it makes scanning through things easier (we can consider another organizational system, that's just what I went with for now.

Thanks for this precision. I wondered how was organised the item. Now I know. I'll change that. And see for `goto` and `exit`.
(And also sorry for the deletion of the branch)

Entomy:
Hey man, it happens, no worries. I didn't have any documentation for these things because typically speaking, you just don't see many contributors in FOSS projects; I'll add this to the `CONTRIBUTING.md` file. Realistically speaking most FOSS is developed by 1-5 people, even for absolutely massive projects. Your contributions are ***very*** appreciated, especially as I'm unable to regularly contribute a lot between work and issues that developed making it difficult to use VS Code itself. *Thank you*

gcluzel:
> On a maybe related note, `goto` and `exit` regressed at some point but I don't know if that happened with this one or elsewhere.

For `goto`, I don't know regression you saw. But in commit e4336d8 (line 620) you delete the `goto` patterns.
For `exit`, I guess that it comes from the same commit, line 450. You now match `#label` which is `"(<<)?((?:\\w|\\d|_)+)(:|>>)"`. Maybe you might add a `?` for the last group, or change for what it was before.

gcluzel:
- I merged my two previous commits and ordered everything in alphabetic order.
- I resolve the when and exit regression in a separate commit.

Entomy:
> For goto, I don't know regression you saw. But in commit e4336d8 (line 620) you delete the goto patterns.

🤦🏻‍♂️

> For **exit**...

Yeah, I'll look at this one myself after this PR is committed.

> Maybe you might add a ? for the last group, or change for what it was before.

Yeah, this was because Ada technically has two different label types, which is... not a decision I agree with. I'll figure out how to handle this well.

gcluzel:
> Yeah, I'll look at this one myself after this PR is committed.

Ok! I revert my last commit and you'll see ;)

setton:
Let's approve/merge this and build on top of it. Thanks a lot for the contribution and the review @Entomy and @gcluzel. If one of you could approve #422, I'll then take care of adding test cases to cover this!

PR #423, opened 2020-04-26
Improve GPR highlighting support

gcluzel:
For my own use I also needed to improve the highlighting support of `.gpr` files.

I added:
- Support for with clause.
- Support for extends keyword in both projects and package.
- Support for package extension.

I also attach an example that should be correctly highlighted for your tests.
```Ada
with "Ada_Drivers_Library/boards/stm32f769_discovery/stm32f769_discovery_sfp.gpr";

project HTTP_Cyclone extends "Ada_Drivers_Library/examples/shared/common/common.gpr" is


   for Runtime ("Ada") use STM32F769_Discovery_sfp'Runtime("Ada");
   for Target use "arm-eabi";
   for Object_Dir use "obj";
   for Create_Missing_Dirs use "True";

   for Source_Dirs use ("src/**");

   package Compiler renames STM32F769_Discovery_SFP.Compiler;

end HTTP_Cyclone;
```
image

Also, last commit changes `"punctuation.definition.string.ada"` for `"punctuation.definition.string.gpr"`. Tell me if it's correct because I'm not sure.

### DIFF
--- a/integration/vscode/ada/syntaxes/gpr.tmLanguage.json
+++ b/integration/vscode/ada/syntaxes/gpr.tmLanguage.json
@@ -100,12 +100,11 @@
 		},
 		"package": {
 			"name": "meta.package.gpr",
-			"begin": "(?i)\\b(package)\\s+((?:\\w|\\d|_)+)\\s+(is)\\b",
+			"begin": "(?i)\\b(package)\\s+((?:\\w|\\d|_)+)\\b",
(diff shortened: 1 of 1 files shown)


Entomy:
@gcluzel that was correct. Classifications should always end in the language being classified. They don't sometimes because i copied similar constructs and forgot to change the language part.

setton:
I agree with @Entomy, makes sense.

I've added the .gpr grammar tester and made an initial test based on this.

Thanks a lot @gcluzel!

PR #426, opened 2020-04-27
T423-090: Display project name for aggregate project tooltips

AnthonyLeonardoGracio:
Since displaying the base name of the entity's declaration file might
not be sufficient in case of aggregate projects.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -1630,7 +1630,9 @@ package body LSP.Ada_Handlers is
             value     => Decl_Text,
             language  => +"ada"));
 
-      --  Append the declaration's location
+      --  Append the declaration's location.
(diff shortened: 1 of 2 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -1643,6 +1645,10 @@ package body LSP.Ada_Handlers is
            (Integer (Decl.Sloc_Range.Start_Column), Min_Width => 1)
          & ")");
 
+      if Self.Project_Tree.Root_Project.Is_Aggregate_Project then
+         Location_Text := Location_Text & " in " & C.Id & " project";
```
setton, line 1649: we should write "in project p" rather than "in p project"

AnthonyLeonardoGracio: Done

PR #433, opened 2020-04-29
Resolve bug in nested declarations

gcluzel:
Before
image
After
image

For @setton if you are interested in adding it in your "nested" test case.
```Ada
procedure Main is
   function D return T
        with X => "Bc";
   
   procedure E is
   begin
      Ada.Text_IO.Put_Line("Hey");
   end E;
begin
 null;
end Main;
```


gcluzel:
I also add a commit to handle more comment lines everywhere. It should solve some problems shown by @raph-amiard with a low cost.

Issue #436, opened 2020-04-29
More highlighting issues


raph-amiard:
![hl_problem_1](https://user-images.githubusercontent.com/478727/80600072-29a38d80-8a2c-11ea-9d69-436f97f12022.png)
![hl_problem_2](https://user-images.githubusercontent.com/478727/80600074-2a3c2400-8a2c-11ea-9385-cf183a40d3de.png)
![hl_problem_3](https://user-images.githubusercontent.com/478727/80600078-2ad4ba80-8a2c-11ea-9225-af1bcc85c239.png)
![hl_problem_4](https://user-images.githubusercontent.com/478727/80600080-2ad4ba80-8a2c-11ea-8d17-eaac183b9750.png)

The last one with the wrongly highlighted comments is the most concerning to me. If you want to reproduce, the file `lsp-ada_handlers.adb` is from the ALS repo.

Honestly at this stage I think we should either:

- Have some kind of dumb "token highlighting" fallback on master, and maybe have a `highlighting-dev` branch where it is deactivated so that remaining holes are still visible, as discussed in #412 
- Completely roll back the new highlighter in favor of the old dumb one, and have the new one on a branch.

This should be complemented by extensive testing one way or another.

Entomy:
@gcluzel since you've been contributing as well I'd like you to weight in.

There's a few ways we can approach this, which I will cover. To address the last statement, @setton has been adding classifier testing which I will also be contributing to when I can, so that is already being worked on. Unfortunantly unit testing for extensions was an afterthought for VS Code, so we've been operating without until I found someone else worked out a mechanism to do so. That being said, there's actually three options for how we can handle this, two that you've mentioned, and other which I mentioned as a late state migration option.

## LSP based classifier

This is the ideal that everything would eventually be based on. It is however, the slowest to develop, as I understand it, with very little documentation. It is required to handle certain parts of Ada's grammar, because of context sensitivity within the language that is impossible to resolve only from syntax. There are TreeSitter based extensions doing this, so the mechanism has been worked out despite little documentation. It is entirely dependent on the state of the LSP to work. I do not beleive it is currently in an adequate state for this yet, however, especially as it crashes quite often and that will stop all classifications being done through the LSP. No matter what, this is the eventual mechanism that will have to be done to handle certain ambiguities.

## Fallback

[...]

PR #438, opened 2020-05-02
Improve highlighter

gcluzel:
*This is a work in progress.*
I have added almost all the constructions that were missing.
The highlighter is much more flexible than the previous one.
I put it here if someone wants to test. Also I'm planing to supply more tests.
It works rather well with correct Ada code, and more or less well with incorrect code.
There are still some cases that need to be improved and probably bugs to solve on my code.
#436 
image
image
image
image


setton:
Judging by the test cases, it looks really good.

gcluzel:
@setton It's not intented to change anything to what you planed in #436 since it's not a proper work and it's not enough tested.

setton:
I'll still merge it before the reorg.

PR #447, opened 2020-05-20
T330-007: update test after changes for Ada 2020

AnthonyLeonardoGracio:


setton:
I suggest rewriting the baseline to avoid the Ada 2020 issue, so it works with multiple compilers... what we're really interested here is testing the diagnostics functionality. Are you happy to have a look @AnthonyLeonardoGracio ?

raph-amiard:
@setton I'm not sure what you're referring to here: The baseline *has* been updated, not the source, and the diagnostic that LAL emits (expecting '[') is due to the new shape of the grammar after the Ada 2020 changes.

The diagnostic is not ideal, but that's the state of LAL diagnostics for now :)

setton:
oh right!

PR #449, opened 2020-05-25
SA14-008: Handle aggregates for completion

AnthonyLeonardoGracio:
Completion snippets that list all the possible shapes (discriminants,
components...) of an aggregate are now proposed when opening a parenthesis
to initialize/assign records.

A new flag has also been added in order to control the number of parameters
at which point named notation should be used in the proposed snippets.

Tests have been adapted since "(" now can trigger completion and thus is
listed in the LSP triggerCharacters config field.

### DIFF
--- a/source/ada/lsp-ada_documents.adb
+++ b/source/ada/lsp-ada_documents.adb
@@ -63,11 +63,12 @@ package body LSP.Ada_Documents is
       return LSP.Messages.Als_Visibility;
 
    function Compute_Completion_Item
-     (Context          : LSP.Ada_Contexts.Context;
-      BD               : Libadalang.Analysis.Basic_Decl;
-      DN               : Libadalang.Analysis.Defining_Name;
-      Snippets_Enabled : Boolean;
-      Is_Dot_Call      : Boolean)
+     (Context                  : LSP.Ada_Contexts.Context;
+      BD                       : Libadalang.Analysis.Basic_Decl;
+      DN                       : Libadalang.Analysis.Defining_Name;
+      Snippets_Enabled         : Boolean;
+      Named_Notation_Threshold : Natural;
+      Is_Dot_Call              : Boolean)
       return LSP.Messages.CompletionItem;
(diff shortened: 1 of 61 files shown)


source/ada/lsp-ada_documents.adb
```diff
@@ -1139,22 +1159,55 @@ package body LSP.Ada_Documents is
 
             Insert_Text := Insert_Text & " (";
 
+            --  Compute number of params to know if named notation should be
+            --  used.
+
+            for Param of Params loop
+               for Id of Param.F_Ids loop
+                  Nb_Params := Nb_Params + 1;
```
setton, line 1167: how about removing the inner loop and do

    Nb_Params := Nb_Params + Param.Length;

?

source/ada/lsp-ada_documents.adb
```diff
@@ -1180,16 +1233,302 @@ package body LSP.Ada_Documents is
...
+                    Optional_InsertTextFormat'
+                      (Is_Set => True,
+                       Value  => Snippet);
+                  Insert_Text := Empty_LSP_String;
+
+                  --  Compute number of components to know if named notation
+                  --  should be used.
+
+                  for Comp of Components loop
+                     for Id of As_Component_Decl (Comp).F_Ids loop
```
setton, line 1469: same comment as above :)

source/ada/lsp-ada_documents.adb
AnthonyLeonardoGracio: Done, thanks for noticing!

source/ada/lsp-ada_documents.adb
AnthonyLeonardoGracio: Done.

reznikmm:
I've rebased this branch over master. Make sure to download it before do any further fix.

AnthonyLeonardoGracio:
> I've rebased this branch over master. Make sure to download it before do any further fix.

Ok thanks!

PR #452, opened 2020-05-27
SB14-037: Add new reference kind "access"

AdrienBoulanger:
Add a reference kind for the following write references:
Foo'Access
Foo'Unrestricted_Access
It allows to filter them from more common write reference like
"assignment" ("Foo :=").
Note: They will still be flagged as write references

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -23,6 +23,7 @@ with Ada.Strings.UTF_Encoding.Wide_Wide_Strings;
 with Ada.Strings.Unbounded; use Ada.Strings.Unbounded;
 with Ada.Directories;
 with Ada.Unchecked_Deallocation;
+with Ada.Wide_Wide_Characters.Handling;
 
 with GNAT.OS_Lib; use GNAT.OS_Lib;
 with GNAT.Strings;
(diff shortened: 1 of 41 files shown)


source/protocol/lsp-messages.ads
```diff
@@ -360,7 +360,7 @@ package LSP.Messages is
    --  }
 
    type AlsReferenceKind is
-     (Simple, Write, Static_Call, Dispatching_Call, Parent, Child);
+     (Simple, Access_Ref, Write, Static_Call, Dispatching_Call, Parent, Child);
```
reznikmm, line 363: Could you please update comments above ( `reference_kinds ALS extension:`) and doc/reference_kinds.md file?

AdrienBoulanger: Done, the naming is a little different because "access" is a keyword in Ada.

source/ada/lsp-ada_handlers.adb
```diff
@@ -1716,6 +1717,29 @@ package body LSP.Ada_Handlers is
...
+         if not Node.Parent.Is_Null and then Node.Parent.Kind in Ada_Name then
+            declare
+               Sibling : constant Ada_Node := Node.Next_Sibling;
+               Text    : constant Wide_Wide_String :=
+                 (if Sibling.Is_Null
+                  then ""
+                  else Ada.Wide_Wide_Characters.Handling.To_Lower
+                    (Sibling.Text));
+            begin
+               return Text = "access" or else Text = "unrestricted_access";
```
reznikmm, line 1748: What's about `X'Unchecked_Access`? Do we need it here? May be `X'Address` also?

AdrienBoulanger: Unchecked_Access looks like a good idea.
For Address looking at the description "Address may be used to return the address of the first element allocated to X", it seems out of scope. If you disagree, I can add it: I have no strong opinion about it.

AnthonyLeonardoGracio: I think that you can add 'Address too, it's used sometimes in bare-metal programming so that will be useful

PR #456, opened 2020-06-04
SC16-001: Implement a custom request to show unit dependencies

AnthonyLeonardoGracio:
A new custom request has been introduced to query the dependencies of a
given unit.

Documentation and a simple test have been added for this custom request.

### DIFF
--- a/source/ada/lsp-ada_contexts.ads
+++ b/source/ada/lsp-ada_contexts.ads
@@ -143,6 +143,10 @@ package LSP.Ada_Contexts is
    function List_Files (Self : Context) return File_Sets.Set;
    --  Return the list of files known to this context.
 
+   function Analysis_Units
+     (Self : Context) return Libadalang.Analysis.Analysis_Unit_Array;
+   --  Return the analysis units for all Ada sources known to this context
+
    procedure Index_File (Self : Context; File : GNATCOLL.VFS.Virtual_File);
    --  Index the given file. This translates to refreshing the Libadalang
    --  Analysis_Unit associated to it.
--- a/source/ada/lsp-ada_documents.ads
+++ b/source/ada/lsp-ada_documents.ads
@@ -25,6 +25,7 @@ with Libadalang.Analysis;
 limited with LSP.Ada_Contexts;
 
 with GNATCOLL.Traces;
+with GNATCOLL.VFS;
 
 package LSP.Ada_Documents is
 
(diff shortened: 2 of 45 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -2022,6 +2026,158 @@ package body LSP.Ada_Handlers is
...
+                  end;
+               end if;
+            end;
+
+         when Libadalang.Common.Ada_Compilation_Unit_List_Range =>
+            declare
+               Comp_Unit_List : constant Compilation_Unit_List :=
+                                As_Compilation_Unit_List (Root);
+            begin
+               for Comp_Unit of Comp_Unit_List loop
```
reznikmm, line 2161: Looks like copy/paste of previous case.

source/ada/lsp-ada_contexts.ads
```diff
@@ -59,6 +59,10 @@ package LSP.Ada_Contexts is
    function Id (Self : Context) return LSP.Types.LSP_String;
    --  Return unique identifier of the context.
 
+   function Analysis_Units
+     (Self : Context) return Libadalang.Analysis.Analysis_Unit_Array;
+   --  Return the analysis units for all Ada sources known to this context
+
```
reznikmm, line 65: Can we move `Get_Importing_Units` into `LSP.Ada_Contexts package` instead of making `Analysis_Units` visible?

source/ada/lsp-ada_documents.ads
```diff
@@ -63,6 +63,12 @@ package LSP.Ada_Documents is
    function Versioned_Identifier
      (Self : Document) return LSP.Messages.VersionedTextDocumentIdentifier;
 
+   function Unit
+     (Self    : Document;
+      Context : LSP.Ada_Contexts.Context)
+      return Libadalang.Analysis.Analysis_Unit;
+   --  Return the analysis unit for Self in the given context
```
reznikmm, line 70: Can we move part of `On_ALS_Show_Dependencies_Request` to `lsp-ada_documents.ads`  instead of make `Unit` visible?

source/protocol/lsp-messages.ads
```diff
@@ -4204,6 +4205,7 @@ package LSP.Messages is
    --	 * @since 3.15.0
    --	 */
    --	serverInfo?: {
+
```
reznikmm, line 4208: Could you drop this an extra empty line to keep original comments intact?

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: No it's not because the type is different (Compilation_Unit vs Compilation_Unit_List). I ddid not find any proper solution to factorize the code unfortunately.

source/ada/lsp-ada_contexts.ads
AnthonyLeonardoGracio: Ok

source/ada/lsp-ada_documents.ads
AnthonyLeonardoGracio: Having a way to retrieve the Unit of a Document seems a good idea to me, why you don't want to make it visible?

source/protocol/lsp-messages.ads
AnthonyLeonardoGracio: Ok

source/ada/lsp-ada_documents.ads
reznikmm: `Document` type provides encapsulation over `Analysis_Unit`. Make `Unit` visible breaks this encapsulation. As result more logic concentrates in `Ada_Handler`, but `Ada_Handler` is expected to by just a protocol adapted.

General architecture of Ada Language server (that is located in `source/ada` directory) consists of:

- `Context_Set` - is set of context, that represents aggregated project. It owns `Context` objects.
- `Context` - a context represents non aggregate project.
- `Document` - a representation of Ada text file. It could be interpreted in several `Context`s if the corresponding Ada file belongs to several projects of an aggregated project.
- `Ada_Handlers.Message_Handler` - handler just a protocol adapter, that uses `Document`, `Context` and `Context_Set` to execute LSP queries.

Put all logic into the `Message_Handler` will result in a mess of code in one huge compilation unit. So, moving code into `Document` helps us to have more clean design.

If a new functionality fits into this architecture, then i would prefer not to make `Unit` visible. Otherwise we should find a new proper architecture.

AnthonyLeonardoGracio: Understood. I tried to follow what you are saying in my last patch.

PR #477, opened 2020-06-29
T626-018 Fix computation of Last parameter

godunko:


source/client/lsp-raw_clients.adb
```diff
@@ -214,19 +214,21 @@ package body LSP.Raw_Clients is
               (1 .. Ada.Streams.Stream_Element_Count (Size))
                 with Import, Address => Slice'Address;
             Last  : Natural;
+
          begin
             Client.Server.Write_Standard_Input
               (Raw, Ada.Streams.Stream_Element_Count (Last));
 
-            if Last = 0 then
```
reznikmm, line 221: The condition `Last = 0` corresponds to API definition:
```ada
   procedure Write_Standard_Input
     (Self : in out Process'Class;
      Data : Ada.Streams.Stream_Element_Array;
      Last : out Ada.Streams.Stream_Element_Offset);
   --  Do nothing if Self.Status /= Running. If no data was written you will
   --  get Standard_Input_Available notification latter.
```

You can expect next Standard_Input_Available only when `Last < Data'Last`.

godunko: Behavior of Last parameter and notification has been clarified in source code.

source/spawn/spawn-processes.ads
```diff
@@ -121,8 +121,10 @@ package Spawn.Processes is
      (Self : in out Process'Class;
       Data : Ada.Streams.Stream_Element_Array;
       Last : out Ada.Streams.Stream_Element_Offset);
-   --  Do nothing if Self.Status /= Running. If no data was written you will
-   --  get Standard_Input_Available notification latter.
+   --  Do nothing if Self.Status /= Running. Last is set to index of the last
+   --  element to be written. Standard_Input_Available notification will be
+   --  called when write operation has been completed. Application is
+   --  responsible to call this subprogram again for remaining data.
```
reznikmm, line 127: I guess, this is an incorrect description of the API, because Standard_Input_Available won't be called on each call of Write_Standard_Input.

Can we keep old semantic of the API (when Write_Standard_Input is called when Last < Data'First, no data was written in other words)? (If not, why?)

godunko: Idea behind of Standard_Input_Available is to allow application to continue previous incomplete write operation. Any value of Last < Data'Last means that operation is incomplete and immediate call to Write_Standard_Input most probably will results in incomplete operation again (most probably with Last = 0). Thus, there is no reasons to do calls immediately, next call should be postponed to "next cycle".

As for me, previous semantic was incorrect. New semantic is compatible with old one, consequence calls will return Last = 0 and will activate Standard_Input_Available notification.

I've added new patch for POSIX implementation.

PR #484, opened 2020-07-06
T318-064: Provide completion for predefined item

AnthonyLeonardoGracio:
We now provide basic completion for keywords, aspects, pragmas and
attributes.

While keyword completion is provided directly by LAL, completion items
for pragmas, aspects and attributes are read through a JSON database.

Currently we only have a database for Ada 2012 aspects, pragmas and
keywords: the plan is to have one database per Ada version, and
determine which database should be used depending on the GNAT install
found in the PATH.

Thus, this is work in progress: we need to add a database per version
and we need to refine the behavior for keyword and attributes'
completion (some LAL bits are missing).

### DIFF
--- a/scripts/predefined_completion.py
+++ b/scripts/predefined_completion.py
@@ -1,0 +1,85 @@
+"""
+This script is used to generate an Ada file containing an hard-coded
+JSON string that contains all the information for predefined completion
+items (pragmas, attributes and aspects).
+
+The script should be laucnhed like this:
+
+   python predefined_completion.py <json_filename>
+
+Where <json_filename> points to a JSON file with the following format:
+
+{
+"PREDEFINED_ADA": {
+  "ASPECT": [
+    {
+      "DOC": "This aspect is equivalent to *note pragma Abstract_State: 1c.",
+      "_id": "0",
+      "_name": "Abstract_State",
+      "_origin": "GNAT RM",
+      "_category": "unknown"
+   },
+   ...
+  "ATTRIBUTE": [
+    {
+      "DOC": "Doc for Abort_Signal pragma",
+      "_id": "0",
+      "_name": "Abort_Signal",
+      "_origin": "GNAT RM",
+      "_category": "variable"
+   },
+   ...
+   "PRAGMA": [
+     {
+       "DOC": "Doc for Abort_Defer pragma",
+       "_id": "0",
+       "_name": "Abort_Defer",
+       "_origin": "GNAT RM"
+   },
+   ...
+}
+"""
+
+import json
+import sys
+import os
+
+
+spec_header = """--  Automatically generated, do not edit.
+
+pragma Style_Checks (Off);
+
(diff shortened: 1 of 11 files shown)


scripts/predefined_completion.py
```diff
@@ -0,0 +1,89 @@
...
+     {
+       "DOC": "Doc for Abort_Defer pragma",
+       "_id": "0",
+       "_name": "Abort_Defer",
+       "_origin": "GNAT RM"
+   },
+   ...
+}
+
+JSON files of this format can be directly retrieved by converting GNAT Studio's
```
AdrienBoulanger, line 42: This comment is related to GS, not sure it applies here

AnthonyLeonardoGracio: Ok, I'll remove it then.

source/ada/lsp-ada_driver.adb
```diff
@@ -47,9 +48,6 @@ with LSP.Stdio_Streams;
 
 procedure LSP.Ada_Driver is
 
-   function Getenv (Var : String) return String;
```
reznikmm, line 50: Why do you move Getenv out of Ada_Driver?
I would prefer keep all environment reading before other threads started.

source/ada/lsp-predefined_completion.adb
```diff
@@ -0,0 +1,214 @@
...
+      when E : others =>
+         Trace.Trace (E);
+         return;
+   end Load_Predefined_Items;
+
+   -----------------------------------
+   -- Load_Predefined_Completion_Db --
+   -----------------------------------
+
+   procedure Load_Predefined_Completion_Db (Trace : Trace_Handle) is
```
reznikmm, line 126: Why do we need to launch the `gnatmake` on each DidChangeConfiguration? I don't see how its output could change

source/ada/lsp-ada_driver.adb
AnthonyLeonardoGracio: Indeed, it's a leftover from a previous solution based on a environment variable to find the JSON databases. I'll put it back in ada-driver.adb

source/ada/lsp-predefined_completion.adb
AnthonyLeonardoGracio: It's not a bad idea actually IMP, since the user may have changed its PATH before reloading his project for instance, no?

reznikmm: AFAIK, there is no way to change environment variables in another process after its start, so GnatStutio can't change PATH in running ALS. The only way is to restart ALS. I guess, ALS doesn't change its environment variables during the execution. Call to `setenv` isn't thread-safe on Linux and should be avoided in ALS.

AnthonyLeonardoGracio: Ok, I'll call this only at startup then.

PR #488, opened 2020-07-08
T707-058 Allow the driver to be run from e3

setton:
... without gnatpython.
Make the necessary changes to function profiles and names, and
adjustments to use Python 3.

Remove our implementation of get_test_list, which doesn't return
a ParsedTest: instead, use the documented API to select test
drivers.

### DIFF
--- a/testsuite/run-tests
+++ b/testsuite/run-tests
@@ -12,11 +12,11 @@ if __name__ == '__main__':
         status = suite.results[k].name
         if status not in ('PASS', 'XFAIL'):
             all_ok = False
-            print "--- {} : {} ---".format(k, status)
+            print("--- {} : {} ---".format(k, status))
             with open(os.path.join(suite.output_dir,
                                    "{}.yaml".format(k)), 'rb') as f:
                 y = f.read()
-                print yaml.safe_load(y).out
+                print(yaml.safe_load(y).out)
 
     if all_ok:
-        print "SUCCESS"
+        print("SUCCESS")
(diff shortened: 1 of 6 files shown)


testsuite/testsuite.py
```diff
@@ -110,40 +110,11 @@ def tear_down(self):
...
-            results = [t for t in results
-                       if any(s in t for s in sublist)]
-
-        logging.info('Found %s tests', len(results))
-        logging.debug('tests:%s', '\n'.join('  ' + r for r in results))
-        return results
+    @property
+    def test_driver_map(self):
+        return {'ada_lsp': JsonTestDriver,
+                'codecs': CodecsTestDriver}
```
pmderodat, line 116: You can now remove the `DRIVERS`, `TEST_SUBDIR` and `TEST_BLACKLIST` class attributes.

testsuite/run-tests
```diff
@@ -12,11 +12,11 @@ if __name__ == '__main__':
...
+            print("--- {} : {} ---".format(k, status))
             with open(os.path.join(suite.output_dir,
                                    "{}.yaml".format(k)), 'rb') as f:
                 y = f.read()
-                print yaml.safe_load(y).out
+                print(yaml.safe_load(y).out)
 
     if all_ok:
-        print "SUCCESS"
+        print("SUCCESS")
```
pmderodat, line 22: The testsuite framework now prints a summary automatically, and the `-E/--show-error-output` option also displays output/logs for failing tests, so you should be able to remove this and just leave:

```python
suite = testsuite.ALSTestsuite(os.path.dirname(__file__))
sys.exit(suite.testsuite_main())
```

setton: This bit doesn't work, `testsuite/run.sh` already has `--show-error-output` but doesn't print the output in case of failing tests.

PR #490, opened 2020-07-10
T318-064: Avoid keyword completion after non-whitespace characters

AnthonyLeonardoGracio:
In order to avoid proposing keywords right after a '(' for instance.

Aborts basic test for keyword completion has also been added.

### DIFF
--- a/source/ada/lsp-ada_documents.adb
+++ b/source/ada/lsp-ada_documents.adb
@@ -268,6 +268,30 @@ package body LSP.Ada_Documents is
       Self.Trace.Trace ("Done applying changes for document " & File);
    end Apply_Changes;
 
+   -----------------
+   -- Get_Text_At --
+   -----------------
+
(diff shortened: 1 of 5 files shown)


source/ada/lsp-ada_documents.adb
```diff
@@ -2391,12 +2414,40 @@ package body LSP.Ada_Documents is
...
+               End_Pos   => LSP.Messages.Position'
+                 (line      => Position.line,
+                  character => Position.character - 2))
+            else
+               "");
+      begin
+         --  Propose keyword completion if we are not within and end label
+         --  and if there is no previous character of if it's a whitespace (we
+         --  don't want to propose keywords after typing '(' to feed subprogram
+         --  parameters for instance).
```
setton, line 2433: some keywords are valid after (, for instance in

     S : String (1 .. 10) := (others => 'a');

PR #494, opened 2020-07-14
T714-019 Code organization

joaopsazevedo:
Extract List_Bodies_Of to LSP.Lal_Utils.

### DIFF
--- a/source/ada/lsp-lal_utils.ads
+++ b/source/ada/lsp-lal_utils.ads
@@ -150,4 +150,16 @@ package LSP.Lal_Utils is
    --  Checks whether the Token's Pattern is delimited by word delimiters
    --  if As_Word is True.
 
+   package Bodies_List is new Ada.Containers.Doubly_Linked_Lists
+     (Defining_Name);
+
+   function List_Bodies_Of
+     (Definition : Defining_Name;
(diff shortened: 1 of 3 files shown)


source/ada/lsp-lal_utils.ads
```diff
@@ -150,4 +150,16 @@ package LSP.Lal_Utils is
    --  Checks whether the Token's Pattern is delimited by word delimiters
    --  if As_Word is True.
 
+   package Bodies_List is new Ada.Containers.Doubly_Linked_Lists
+     (Defining_Name);
+
+   function List_Bodies_Of
```
reznikmm, line 156: Could you please turn the function into a procedure here?
 I guess it will be a bit easier to understand then a function with `out` parameter.

setton: Ha! Joao started with a procedure but it was me who suggested to make it a function, with homogeneity with `Resolve_Name` and `Find_All_Calls` in this package, which are already implemented as functions with `out` boolean.

I suggest we leave this refactoring for a later stage: we should make the `Imprecise` parameters `in out` everywhere, so that clients can just do
    Result := The_Query (..., Imprecise);
rather than having to do (currently):
    Result := The_Query (..., This_Imprecise);
    Imprecise := Imprecise or This_Imprecise;

PR #499, opened 2020-07-15
Add invisible symbols to completion request.

reznikmm:


source/ada/lsp-ada_handlers.adb
```diff
@@ -2951,13 +2979,42 @@ package body LSP.Ada_Handlers is
...
+               Doc.Get_Any_Symbol_Completion
+                 (Context.all, Canonical_Prefix, 10, Invisible);
+            end loop;
+
+            Invisible.Write_Completions (10, Result);
+         end if;
+      end;
+
+      Response.result.items.Move (Source => Result.Completion_List);
+      Response.result.isIncomplete := Result.Is_Incomplete;
```
AnthonyLeonardoGracio, line 3017: Noe that we don't support incomplete results on GS side for now

PR #516, opened 2020-08-06
T805-014 Replace Defining_Name equality operator

reznikmm:
in global name cache as workaround of an issue from T806_020


reznikmm:
@setton I have added a testcase and a comment. I've also reverted Hash function and changed Is_Equal to use LAL."=" to overcome a possible compiler bug (Thank @Roldak for the investigation and idea!).

PR #534, opened 2020-09-01
T722-025 LAL Tools Refactor Imports Integration

joaopsazevedo:
New codeAction command and handler to perform refactoring of
"with" clauses and prefixes.
Test case to prove that the protocol messages are correct. Therefore,
the test project has an empty runtime so that the test runs faster.

### DIFF
--- a/testsuite/ada_lsp/refactor_imports_commands/refactor_imports_commands.json
+++ b/testsuite/ada_lsp/refactor_imports_commands/refactor_imports_commands.json
@@ -1,0 +1,361 @@
+[
+    {
+        "comment": [
+            "This test checks if imports (with clause + prefix) are being suggested to unresolved names."
+        ]
+    },
+    {
+        "start": {
+            "cmd": [
+                "${ALS}"
+            ]
+        }
+    },
+    {
+        "send": {
+            "request": {
+                "params": {
+                    "processId": 13950,
+                    "capabilities": {
+                        "workspace": {
+                            "applyEdit": false
+                        }
+                    },
+                    "rootUri": "$URI{.}"
+                },
+                "jsonrpc": "2.0",
+                "id": 1,
+                "method": "initialize"
+            },
+            "wait": [
+                {
+                    "id": 1,
+                    "result": {
+                        "capabilities": {
+                            "textDocumentSync": 2,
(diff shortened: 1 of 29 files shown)


testsuite/ada_lsp/refactor_imports_commands/refactor_imports_commands.json
```diff
@@ -0,0 +1,342 @@
...
+                            "projectFile": "default.gpr",
+                            "scenarioVariables": {},
+                            "defaultCharset": "ISO-8859-1"
+                        }
+                    }
+                },
+                "jsonrpc": "2.0",
+                "method": "workspace/didChangeConfiguration"
+            },
+            "wait": []
```
reznikmm, line 103: I suggest you to add this:

            "wait": [
                {
                    "jsonrpc": "2.0",
                    "id": 2,
                    "method": "window/workDoneProgress/create",
                    "params": {
                        "token": "<ANY>"
                    }
                },  {
                    "jsonrpc": "2.0",
                    "method": "$/progress",
                    "params": {
                      "token": "<ANY>",
                      "value": {
                        "kind": "end"
                      }
                    }
                }
            ]
        }

This will wait until ALS finish the indexing, so you will be sure there is no other indexing requests from the ALS.

testsuite/ada_lsp/refactor_imports_commands/refactor_imports_commands.json
```diff
@@ -0,0 +1,342 @@
+[
+    {
+        "comment": [
+            "test automatically generated"
```
reznikmm, line 4: Please update the comment there.

testsuite/ada_lsp/refactor_imports_commands/refactor_imports_commands.json
joaopsazevedo: Agree. Exactly what I was looking for.

testsuite/ada_lsp/refactor_imports_commands/refactor_imports_commands.json
joaopsazevedo: Agree and updated.

reznikmm (approved):

PR #544, opened 2020-09-13
Ensure scripts are compatible with both Python 2 and 3

simonjwright:
* scripts/generate.py: All files that are to be opened for write and
      are to contain text must be opened "w", not "wb".


setton (approved):

setton:
Thank you @simonjwright!

PR #545, opened 2020-09-13
Generate client libraries

simonjwright:
* Makefile (DESTDIR): Defaulted to $prefix, which is defaulted to
      /usr/local.
    (LIBRARY_FLAGS): Merge the two versions.
    (all): Additionally build lsp_client.gpr and lsp_client_glib.gpr.
    (install): Don't install tester.gpr or codec_test.gpr. Instead,
      install lsp_client.gpr and lsp_client_glib.gpr.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,10 +19,13 @@ TD=testsuite/ada_lsp
 GPRBUILD=gprbuild -j0
 
 # Installation directory
-DESTDIR=
+prefix ?= /usr/local
+ifeq ($(DESTDIR),)
+  DESTDIR=$(prefix)
+endif
 
 # Library type
-LIBRARY_TYPE=relocatable
+LIBRARY_TYPE?=relocatable
 
 # Build mode (dev or prod)
 BUILD_MODE=dev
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -77,6 +76,10 @@ all: coverage-instrument
 	$(GPRBUILD) -P gnat/lsp_server.gpr -p $(COVERAGE_BUILD_FLAGS) \
 		-XVERSION=$(TRAVIS_TAG)
 	$(GPRBUILD) -P gnat/codec_test.gpr -p $(COVERAGE_BUILD_FLAGS)
+	$(GPRBUILD) -P gnat/lsp_client.gpr -p $(COVERAGE_BUILD_FLAGS) \
+		-XVERSION=$(TRAVIS_TAG)
+	$(GPRBUILD) -P gnat/lsp_client_glib.gpr -p $(COVERAGE_BUILD_FLAGS) \
```
setton, line 81: We'd rather not build the _glib client libraries systematically - they are really only useful for integration in GNAT Studio and we might move the code there in the future, keeping the Ada Language Server buildable/testable without the dependency on GtkAda.

setton (changes requested):

Makefile
simonjwright: OK, change pushed.
Noted that the Makefile should contain something to ensure the `generate` target gets run.
Noted that the README should really be revised.

setton: Thank you for the change, and I do agree the README is due for a refresh.

Makefile
```diff
@@ -99,8 +102,14 @@ endif
 install:
 	gprinstall -f -P gnat/lsp_server.gpr -p -r --mode=usage \
 		--prefix=$(DESTDIR) $(LIBRARY_FLAGS)
-	gprinstall -f -P gnat/tester.gpr -p --prefix=$(DESTDIR) $(LIBRARY_FLAGS)
-	gprinstall -f -P gnat/codec_test.gpr -p --prefix=$(DESTDIR) $(LIBRARY_FLAGS)
+	gprinstall -f -P gnat/lsp_client.gpr -p -r	\
```
setton, line 102: The install of lsp_client_glib.gpr should also be left out of this section.

simonjwright: The first commit did install lsp_client_glib.gpr, this one doesn’t - or have I messed something?
The clean: target should clean up lsp_client.gpr (and lsp_client_glib.gpr).
Should there be a non-default, temporary, option to build & install lsp_client_glib.gpr?

setton: My bad! I was reviewing the first commit indeed.

setton: I think we're okay without specific handling of the *glib bits.

setton (approved):

PR #547, opened 2020-09-16
T122-009: Add a 'displayTypeHierarchy' configuration flag

AnthonyLeonardoGracio:
This flag allows to customize the behavior of the navigation requests
(e.g: textDocument/definition) regarding the listing of overriding/overridden
subprograms.

A dedicated setting has been added to the VS Code extension.

We don't list overridding/overridden subprograms on textDocument/declaration
anymore, no matter the value of the preference: if we are on a body, we should
always go to the corresponding declaration. The user can then navigate through
the hierarchy from the declaration.

Tests have been adjusted: more tests will come once we fully agree on the
behavior.


integration/vscode/ada/package.json
```diff
@@ -138,6 +138,23 @@
                         "default": "iso-8859-1",
                         "description": "The character set to use while reading files from disk."
                     },
+                    "ada.displayTypeHierarchy": {
```
setton, line 141: it should not be called `displayTypeHierarchy`, since this is not really a hierarchy of types. How about `displayMethodAncestryOnNavigation` ? a bit long but we don't care, this is not shown to standard users...

scripts/io_gen.py
```diff
@@ -11,6 +11,7 @@
     'Span',
     'CodeActionKind',
     'AlsReferenceKind',
+    'AlsDisplayTypeHierarchyPolicy',
```
setton, line 14: (if you agree with changing the name of the flag in json, this should be changed accordingly - same for the Ada code)

setton (changes requested):

integration/vscode/ada/package.json
AnthonyLeonardoGracio: Agreed, I will change it

scripts/io_gen.py
AnthonyLeonardoGracio: Done

setton (approved):

PR #555, opened 2020-10-02
T812-022: Imprecise results for text/Definition

AdrienBoulanger:
Better handling to navigate between the declaration/body/ref
when the profiles are slightly different (common case:
the user is modifying the spec profile and then wants to
navigate to the body)

Add a test.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -1300,7 +1300,7 @@ package body LSP.Ada_Handlers is
             end if;
          else  --  If we are on a defining_name already
             Other_Part := Laltools.Common.Find_Next_Part
-              (Definition, Self.Trace);
+              (Definition, Self.Trace, Imprecise_Fallback => True);
 
             Decl_For_Find_Overrides := Definition.P_Basic_Decl;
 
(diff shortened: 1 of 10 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -1300,7 +1300,7 @@ package body LSP.Ada_Handlers is
             end if;
          else  --  If we are on a defining_name already
             Other_Part := Laltools.Common.Find_Next_Part
-              (Definition, Self.Trace);
+              (Definition, Self.Trace, Imprecise_Fallback => True);
```
setton, line 1303: This change breaks the fact that the user gets warned when the results are not precise.

I suggest doing the following:

  - call `Find_Next_Part` with `Imprecise_Fallback => False`
  - if `Other_Part = No_Defining_Name`, set the boolean `Imprecise` to True and re-try `Find_Next_Part` with `Imprecise_Fallback => True`

source/ada/lsp-ada_handlers.adb
```diff
@@ -1318,7 +1318,7 @@ package body LSP.Ada_Handlers is
             if Other_Part = No_Defining_Name then
                --  No next part is found. Check first defining name
                Other_Part := Laltools.Common.Find_Canonical_Part
-                 (Definition, Self.Trace);
+                 (Definition, Self.Trace, Imprecise_Fallback => True);
```
setton, line 1321: Same remark as above: add a 2-step mechanism to make sure we set `Imprecise` when `Imprecise_Fallback` is needed.

setton (changes requested):

source/ada/lsp-ada_handlers.adb
AdrienBoulanger: I will be slower: it is not recommended to call Find_Next_Part 2 times. Furthermore I would like to remove this Imprecise warning altogether.

PR #557, opened 2020-10-07
T928-015: Sanitize the result of the LCS algorithm

AdrienBoulanger:
The LCS algorithm generates a diff between lines: it will return
a wrong result if the buffer doesn't finish by a EOL symbol.
=> Add a defensive function to set all the wrong location to
the end of the buffer.

### DIFF
--- a/source/ada/lsp-ada_documents.adb
+++ b/source/ada/lsp-ada_documents.adb
@@ -493,6 +493,26 @@ package body LSP.Ada_Documents is
          Add (Old_Line_Number);
          Free (LCS);
 
+         --  Handle the edge case where the last location of
+         --  the edit is trying to affect a non existent line.
+         --  The edits are ordered so we only need to check the last one.
+         if not (Integer (Edit.Last_Element.span.last.line)
+                 in Self.Line_To_Index.First_Index ..
+                   Self.Line_To_Index.Last_Index)
+         then
+            declare
+               Element : LSP.Messages.TextEdit := Edit.Last_Element;
+            begin
+               --  Replace the wrong location by the end of the buffer
+               Element.span.last :=
+                 (line => Line_Number
+                    (Old_Lines.Last_Index - Old_Lines.First_Index),
+                  character =>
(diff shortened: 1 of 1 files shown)


source/ada/lsp-ada_documents.adb
```diff
@@ -234,6 +234,7 @@ package body LSP.Ada_Documents is
             end;
          else
             Self.Text := Change.text;
+            Self.Trace.Trace ("Hello There");
```
ogorodnik, line 237: I guess this should be removed.

source/ada/lsp-ada_documents.adb
```diff
@@ -492,6 +523,7 @@ package body LSP.Ada_Documents is
 
          Add (Old_Line_Number);
          Free (LCS);
+         Sanitize;
```
ogorodnik, line 526: Instead of Sanitize I would add the following code:

         if Self.Line_To_Index.Find
           (Integer (Edit.Last_Element.span.last.line)) = No_Element
         then
            declare
               Element : LSP.Messages.TextEdit := Edit.Last_Element;
            begin
               Element.span.last :=
                 (line => Line_Number
                    (Old_Lines.Last_Index - Old_Lines.First_Index),
                  character =>
                    UTF_16_Index (Length (Old_Lines.Last_Element)));
               Edit.Replace_Element (Edit.Last, Element);
            end;
         end if;

In this case we will not raise an exception and also we don't need to process all elements because we can have only one TextEdit for the last line in the end of the list

ogorodnik (changes requested):

source/ada/lsp-ada_documents.adb
AdrienBoulanger: Maybe? ;)

source/ada/lsp-ada_documents.adb
AdrienBoulanger: I like the idea

ogorodnik (approved):

PR #558, opened 2020-10-07
TA06-033 Do not emit a warning 'multiple projects' too early

setton:
Emit this warning only once we have received didChangeConfiguration
and the project is still ambiguous at that point.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -179,10 +179,13 @@ package body LSP.Ada_Handlers is
 
    procedure Ensure_Project_Loaded
      (Self : access Message_Handler;
+      Warn : Boolean;
       Root : LSP.Types.LSP_String := LSP.Types.Empty_LSP_String);
    --  This function makes sure that the contexts in Self are properly
    --  initialized and a project is loaded. If they are not initialized,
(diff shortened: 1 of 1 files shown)


reznikmm (commented):
How do you get this warning?

As for me this is important to let users know that ALS is unable to find a right project.

setton:
> How do you get this warning?
> 
> As for me this is important to let users know that ALS is unable to find a right project.

The warning appears if didChangeConfiguration arrives and no project was specified.

reznikmm:
But GNAT Studio *must* send the project in didChangeConfiguration, right? So this should be fixed on GNAT Studio side, I guess.

setton:
> But GNAT Studio _must_ send the project in didChangeConfiguration, right? So this should be fixed on GNAT Studio side, I guess.

I'm not sure I follow... GNAT Studio is always working on a .gpr, so it always passes a .gpr to the ALS: there is never a reason to warn about "multiple projects found" for GNAT Studio. This needs to be sent to users of VScode who have not specified a project: if the ALS cannot find a project by the time didChangeConfiguration completes, then a message is sent back to VS Code.

setton:
In other words, this fixes a bug whereby users of GNAT Studio were notified of "multiple projects" even though they shouldn't.

PR #568, opened 2020-10-19
Topic/ta14 018

godunko:


source/spawn/spawn-processes-windows.adb
```diff
@@ -239,56 +244,178 @@ package body Spawn.Processes.Windows is
...
-            Success := False;
-            return;
-         end if;
+         --  Create named pipe and server handle
+
+         loop
+            declare
+               Pipe_Count_Image : constant String :=
+                 Interfaces.Unsigned_32'Image (Pipe_Count);
+               --  Pipe_Name        : constant Interfaces.C.char_array :=
```
setton, line 285: commented out code, should be removed

source/spawn/spawn-processes-windows.adb
```diff
@@ -239,56 +244,178 @@ package body Spawn.Processes.Windows is
...
+                     Parent_Handle := System.Win32.INVALID_HANDLE_VALUE;
+                     Success       := False;
+
+                     return;
+                  end if;
+               end;
+            end;
+         end loop;
+
+         --  Open named pipe to create clinet handle
```
setton, line 341: *client

setton (changes requested):

source/spawn/spawn-processes-windows.adb
```diff
@@ -239,56 +244,176 @@ package body Spawn.Processes.Windows is
...
-            dwFlagsAndAttributes  => System.Win32.FILE_ATTRIBUTE_NORMAL,
+            dwFlagsAndAttributes  => Windows_API.FILE_FLAG_OVERLAPPED,
             hTemplateFile         => 0);
 
          if Child_Handle = System.Win32.INVALID_HANDLE_VALUE then
             Self.Listener.Error_Occurred (Integer (System.Win32.GetLastError));
-            Success := False;
+
+            Interfaces.C.Strings.Free (Pipe_Name);
+            Ignore (System.Win32.CloseHandle (Parent_Handle));
```
setton, line 354: I feel like we shouldn't ignore the result here and emit a trace in case it's False

setton (approved):

source/spawn/spawn-processes-windows.adb
godunko: It is not necessary here, handle is valid one and not created by some specific functions which requires specific cleanup

PR #573, opened 2020-11-05
TA09-019 Add support for filesystem monitoring

setton:
This adds support for monitoring the filesystem for changes
that might occur in project directories, outside of the IDE.

Add support for `workspace/didChangeWatchedFiles`.

Add an optional dependency on ada_libfswatch.gpr, and document
it. Add support in the travis CI builder to generate it on Linux.
Add a stub project to make the ALS useable without this dependency.

Rework the way files are indexed, to allow reindexing for any
file that might change on disk. In particular, no longer store
in LSP.Ada_Contexts a reference to the "Last_Indexed" file, which
prevented indexing of new files altogether.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -122,11 +122,15 @@ package body LSP.Ada_Handlers is
    procedure Release_Project_Info (Self : access Message_Handler);
    --  Release the memory associated to project information in Self
 
+   function Contexts_For_File
+     (Self : access Message_Handler;
+      File : Virtual_File)
+      return LSP.Ada_Context_Sets.Context_Lists.List;
    function Contexts_For_URI
      (Self : access Message_Handler;
       URI  : LSP.Messages.DocumentUri)
       return LSP.Ada_Context_Sets.Context_Lists.List;
-   --  Return a list of contexts that are suitable for the given URI:
+   --  Return a list of contexts that are suitable for the given File/URI:
    --  a list of all contexts where the file is known to be part of the
    --  project tree, or is a runtime file for this project. If the file
    --  is not known to any project, return an empty list.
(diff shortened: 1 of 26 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -3195,40 +3270,42 @@ package body LSP.Ada_Handlers is
...
-                  Current_Percent := (Index * 100) / Total;
-                  --  If the value of the indexing increased by at least one
-                  --  percent, emit one progress report.
-                  if Current_Percent > Last_Percent then
-                     Emit_Progress_Report (Current_Percent);
-                     Last_Percent := Current_Percent;
-                  end if;
+      while not Self.Files_To_Index.Is_Empty loop
+         declare
+            Cursor : File_Sets.Cursor := Self.Files_To_Index.First;
```
reznikmm, line 3275: Minor note: We could get rid of Cursor altogether using

    File := Self.Files_To_Index.First_Element;
    Self.Files_To_Index.Delete_First;

source/ada/lsp-ada_handlers.ads
```diff
@@ -135,28 +135,31 @@ private
...
-      --        or a requests, it looks at whether Indexing_Required is True,
-      --        if it is, it runs Indexing_Required
-      --
-      --      * Indexing_Required is set to False when Index_Files has
-      --        completed
+      --        or a requests, it looks at whether Files_To_Index contains
+      --        files; if it does, it runs Index_Files
+
+      Total_Files_Indexed  : Natural := 0;
+      Total_Files_To_Index : Positive := 1;
```
reznikmm, line 159: I guess it's safe to make `Total_Files_To_Index : Natural := 0;`, because we don't calculate percentage if there is no files to index.

reznikmm (approved):
I would note, that LSP client won't send us `workspace/didChangeWatchedFiles` until we do dynamic registration for this notification.

I have 2 minor comments, but fill free to merge this patch.

setton:
Thank you for the review @reznikmm - I will merge and address your comments with a new PR.

source/ada/lsp-ada_handlers.adb
setton: ah there is no First_Element or Delete_First for a Hashed_Set!

PR #575, opened 2020-11-12
TA09-019 filesystem monitoring: better handle exiting

setton:
Prevent potential call to Stop_Monitor in case the monitor
was never started.


AnthonyLeonardoGracio (approved):
The build failure does not seem related with yoru patch

AnthonyLeonardoGracio:
The build failure is due to a sync pb: LAL has just been bumped today. Feel free to push!

PR #577, opened 2020-11-16
TA09-019 Add a trace to control filesystem monitoring

setton:
Turn it off by default for now.

### DIFF
--- a/source/server/lsp-servers.adb
+++ b/source/server/lsp-servers.adb
@@ -50,6 +50,10 @@ package body LSP.Servers is
    New_Line : constant String :=
      (Ada.Characters.Latin_1.CR, Ada.Characters.Latin_1.LF);
 
+   Filesystem_Monitoring_Trace : constant GNATCOLL.Traces.Trace_Handle :=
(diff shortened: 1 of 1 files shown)


source/server/lsp-servers.adb
```diff
@@ -50,6 +50,10 @@ package body LSP.Servers is
    New_Line : constant String :=
      (Ada.Characters.Latin_1.CR, Ada.Characters.Latin_1.LF);
 
+   Filesystem_Monitoring_Trace : constant GNATCOLL.Traces.Trace_Handle :=
```
reznikmm, line 53: Can we use a configuration item instead of the trace? It will be easier for users of VS Code to change it. Also it would be mo visible to VS Code users.

setton: I'll do this later: now we need to urgently deactivate this so that we can get GS wavefronts

AnthonyLeonardoGracio: Agreed

AnthonyLeonardoGracio (approved):

PR #584, opened 2020-11-26
Update ALS souces after renaming of Each_Stream_Element

reznikmm:
in VSS Stream_Element_Buffer.

### DIFF
--- a/source/server/lsp-servers.adb
+++ b/source/server/lsp-servers.adb
@@ -41,6 +41,7 @@ with Libadalang.Common;         use Libadalang.Common;
 
 with VSS.JSON.Streams.Readers.Simple;
 with VSS.Stream_Element_Buffers;
+with VSS.Stream_Element_Buffers.Conversions;
 with VSS.Strings.Conversions;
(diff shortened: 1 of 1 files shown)


source/server/lsp-servers.adb
```diff
@@ -993,9 +993,9 @@ package body LSP.Servers is
...
 
             begin
-               for E in Vector.Each_Stream_Element loop
+               for E of Vector loop
                   Last := Last + 1;
-                  Aux (Last) := Character'Val (E.Element);
+                  Aux (Last) := Character'Val (E);
                end loop;
 
                Server.Out_Trace.Trace (Aux (Aux'First .. Last));
```
godunko, line 1001: I suppose result of VSS.Stream_Element_Vectors.Conversions.Unchecked_To_String function can be passed directly here, without intermediate buffer.

Use of Last looks strange here, isn't Last = Vector.Length here?

reznikmm: A good idea, thanks

godunko (approved):

PR #592, opened 2020-12-14
(spawn.gpr): In production mode, disable warning-as-errors

setton:


gnat/spawn.gpr
```diff
@@ -20,12 +20,21 @@ project Spawn is
    type OS_API_Kind is ("unix", "osx", "Windows_NT");
    OS_API : OS_API_Kind := external ("OS", "unix");
 
-   --  By default, treat warnings as errors in dev mode, but not in prod
-   --  mode. Let users override this default using the SPAWN_WARN_ERRORS
-   --  environment variable.
+   type Build_Type is ("Debug", "Production");
+   Build : Build_Type := External ("BUILD", "Debug");
```
reznikmm, line 24: The ALS uses LAL scenario variable `BUILD_MODE` with `prod` and `dev` values.

setton: we need to homogeneize this because GNAT Studio uses `BUILD` with `Production` and `Debug`... Today the issue appeared in GNAT Studio. Let's change LAL ?

Fabien-Chouteau: You must avoid using only `BUILD` as external to avoid clashes (see S709-037).
It should be replaced by `SPAWN_BUILD_MODE`

PR #609, opened 2021-02-09
U203-003 Don't raise exception on unknown request id...

godunko:
... and pass internal error to the caller for processing.


reznikmm (approved):
Perhaps we can just return error as a string instead of returning a boolean.

PR #620, opened 2021-02-23
S314-015 Add subprogram signature refactoring tools

joaopsazevedo:


reznikmm:
I guess, you should redo this PR to `master` branch.

source/ada/lsp-ada_handlers.adb
```diff
@@ -676,6 +680,17 @@ package body LSP.Ada_Handlers is
         (Is_Set => True,
          Value => (Is_Boolean => True, Bool => True));
 
+      if Value.capabilities.workspace.workspaceEdit.documentChanges.Is_Set then
```
reznikmm, line 683: This `if` could be rewritten a bit shorter using `True` value for `Optional_Boolean` type:
```ada
Self.Versioned_Documents := 
  Value.capabilities.workspace.workspaceEdit.documentChanges = True;
```

source/ada/lsp-ada_handlers-refactor_change_parameter_mode.adb
```diff
@@ -0,0 +1,415 @@
...
+         else
+            return Ada_Mode_Default;
+         end if;
+      end Value;
+
+   begin
+      Target_Subp := Resolve_Name_Precisely (Get_Node_As_Name (Node));
+
+      if Target_Subp.Is_Null then
+         raise Program_Error with
```
reznikmm, line 342: Can we set `Error` parameter instead of raising a `Program_Error` here?

source/ada/lsp-ada_handlers-refactor_move_parameter.adb
```diff
@@ -0,0 +1,318 @@
...
+   is
+      Pointer : LSP.Commands.Command_Pointer;
+      Code_Action : LSP.Messages.CodeAction;
+      Where       : constant LSP.Messages.Location :=
+        LSP.Lal_Utils.Get_Node_Location (Target_Subp.P_Defining_Name.F_Name);
+
+      function Image
+        (D : Move_Direction_Type)
+         return Langkit_Support.Text.Text_Type
+      is ((if D = Right then "right" else "left"));
```
reznikmm, line 57: Is it `right` and `left` good for moving "direction" description?
Personally, I almost never have two parameters in line, but one under another.
Could more general (`before`/`after` or `forward`/`backward`) be better for this?

source/ada/lsp-ada_handlers-refactor_change_parameter_mode.adb
```diff
@@ -0,0 +1,415 @@
...
+      --  a WorkspaceEdit.
+
+      function Value (M : LSP.Types.LSP_String) return Ada_Mode;
+      --  Converts an LSP_String to an Ada_Mode
+
+      -----------------------
+      -- Get_Text_Document --
+      -----------------------
+
+      function Get_Text_Document
```
reznikmm, line 296: To avoid code duplication we can have something like `Document_Version` method in the Message_Handler, I think.

source/ada/lsp-ada_handlers-refactor_move_parameter.adb
joaopsazevedo: Agree.

source/ada/lsp-ada_handlers-refactor_change_parameter_mode.adb
joaopsazevedo: Agree.

source/ada/lsp-ada_handlers-refactor_change_parameter_mode.adb
joaopsazevedo: Agree, I'll add a `Get_Open_Document_Version` method.
`Message_Handler` implements `Document_Provider` interface.
Since `VersionedTextDocumentIdentifier` has a `Nullable_Number`, I'm adding this method to the `Document_Provider` interface too and `Document_Provider`s that do not have versioned documents will return a null number.

reznikmm (approved):

PR #623, opened 2021-03-09
U308-004 U305-040 Minor update on exception handling

setton:
Catch Predicondition_Failure in the same way we intercept
Property_Error, ie assuming that this can happen in the normal
flow when working on invalid code, and not report them to the
IDE.

Systematically store the symbolic backtrace of exceptions in the
log.

Factorize the code to do this for both notifications and requests.

### DIFF
--- a/source/server/lsp-servers.adb
+++ b/source/server/lsp-servers.adb
@@ -1101,6 +1101,13 @@ package body LSP.Servers is
          Server       : not null LSP.Server_Backends.Server_Backend_Access);
       --  Initializes internal data structures
 
+      function Process_Exception
+        (Message : String; E : Exception_Occurrence) return Boolean;
+      --  Process the exception: depending on the type of exception, send a
+      --  response to the user or store it in the log.
+      --  Return True if the exception is known to be raised by Libadalang
+      --  when processing invalid Ada code.
+
       procedure Process_Message (Message : in out Message_Access);
 
       ----------------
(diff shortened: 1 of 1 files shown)


source/server/lsp-servers.adb
```diff
@@ -1130,14 +1159,24 @@ package body LSP.Servers is
...
-            LSP.Messages.Server_Notifications.Server_Notification'Class
-              (Message.all).Visit (Notif_Handler);
-            Server_Backend.After_Work (Message.all);
+            begin
+               Server_Backend.Before_Work (Message.all);
+               LSP.Messages.Server_Notifications.Server_Notification'Class
+                 (Message.all).Visit (Notif_Handler);
+               Server_Backend.After_Work (Message.all);
+            exception
+               when E : others =>
```
godunko, line 1168: Why not to use Ada here?

when Property_Error | Precondition_Failure =>
  do something
  
when others =>
 do another

I found it more understandable then use of some magic subprogram to do something even more magic with exception identifiers.

source/server/lsp-servers.adb
```diff
@@ -1173,33 +1212,24 @@ package body LSP.Servers is
             --  was raised when processing the request.
             --
             when E : others =>
-               Send_Exception_Response
-                 (Server.all, E,
-                  Ada.Tags.External_Tag (Message'Tag), Request.id);
+               --  Trace the exception
```
godunko, line 1215: Same is here

source/server/lsp-servers.adb
setton: I don't think there is anything "magic" in the exception identifiers, this comes from http://www.ada-auth.org/standards/2xaarm/html/AA-11-4-1.html, paragraphs 7, 9, and 11 ;-) ;-)

The reason I'm using a separate function is that I would like to factorize the code between the two cases.

yannickmoy (approved):

PR #630, opened 2021-03-12
U308-004 When receiving an exception, always send a response

setton:
Add code to send an empty response when the exception is due
to "invalid Ada": we don't want to send a backtrace in this case,
but clients are still interested to get the response to the request.

### DIFF
--- a/source/ada/lsp-error_decorators.adb
+++ b/source/ada/lsp-error_decorators.adb
@@ -14,6 +14,8 @@
 -- COPYING3.  If not, go to http://www.gnu.org/licenses for a complete copy --
 -- of the license.                                                          --
 ------------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


reznikmm (changes requested):
I think we should use `LSP.Error_Decorators` instead. Each request may have distinct format of an empty `result`. That's why Error_Decorator is invented.

setton:
Good point, I’ll take a look.

setton:
> I think we should use `LSP.Error_Decorators` instead. Each request may have distinct format of an empty `result`. That's why Error_Decorator is invented.
> Also I'm not sure that sending `Is_Error=>True` without any explanation in the `error` field is a good idea.

Done, moved the logic to Error_Decorators. It works.

reznikmm (approved):

PR #635, opened 2021-03-15
Update gitpod Dockerfile to install GNAT Community 2020

reznikmm:


setton (approved):
approved but how about adding `.vscode/settings.json` to indicate a project file as well?

PR #640, opened 2021-03-17
U304-005 Add basic support for preprocessor directives

setton:


setton:
@reznikmm this is the follow up on #626 which addresses your comments:

  - we no longer change the line terminators passed to LAL
  - we no longer send the "#end" directive to LAL

Your comment regarding the fact that reformatting won't work for preprocessed code still stands - but this didn't work either before this patch (LAL simply ignores all preprocessing directives) and the situation is much improved for navigation in the presence of preprocessing.

source/ada/lsp-unit_providers.adb
```diff
@@ -0,0 +1,207 @@
...
+with LSP.Preprocessor; use LSP.Preprocessor;
+
+package body LSP.Unit_Providers is
+
+   package US renames Ada.Strings.Unbounded;
+   use type US.Unbounded_String;
+
+   type Project_Unit_Provider is new LAL.Unit_Provider_Interface with record
+      Tree             : Prj.Project_Tree_Access;
+      Projects         : Prj.Project_Array_Access;
```
reznikmm, line 33: Currently this provider always uses a single project. Why not replace the array with a single project?

source/ada/lsp-unit_providers.ads
```diff
@@ -0,0 +1,70 @@
...
+
+   function Create_Project_Unit_Provider
+     (Tree             : Prj.Project_Tree_Access;
+      Project          : Prj.Project_Type := Prj.No_Project;
+      Env              : Prj.Project_Environment_Access;
+      Is_Project_Owner : Boolean := True)
+      return LAL.Unit_Provider_Reference;
+   --  Likewise, but create only one unit provider.
+   --
+   --  If a non-null Project is given, use it to provide units. Raise an
```
reznikmm, line 50: null project isn't supported according to the function body, right?

source/ada/lsp-unit_providers.ads
```diff
@@ -0,0 +1,70 @@
...
+
+package LSP.Unit_Providers is
+
+   package LAL renames Libadalang.Analysis;
+   package Prj renames GNATCOLL.Projects;
+
+   Trace : constant GNATCOLL.Traces.Trace_Handle := GNATCOLL.Traces.Create
+     ("LIBADALANG.PROJECT_PROVIDER", GNATCOLL.Traces.From_Config);
+
+   type Provider_And_Projects is record
```
reznikmm, line 35: Is it used somewhere?

reznikmm:
Shall we make "document format" request do nothing/raise error in case of documents with preprocessor directives? Because we know that it won't work right anyway.

source/ada/lsp-unit_providers.adb
```diff
@@ -0,0 +1,207 @@
...
+      Reparse     : Boolean := False) return LAL.Analysis_Unit'Class;
+
+   overriding procedure Release (Provider : in out Project_Unit_Provider);
+   ----------------------------------
+   -- Create_Project_Unit_Provider --
+   ----------------------------------
+
+   function Create_Project_Unit_Provider
+     (Tree             : Prj.Project_Tree_Access;
+      Project          : Prj.Project_Type := Prj.No_Project;
```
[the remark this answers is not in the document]
setton: The reason these are here is that this package is a straight copy of the Libadalang unit provider, the only change being where the file is read from disk - see the doc in the .ads.

setton:
> Shall we make "document format" request do nothing/raise error in case of documents with preprocessor directives? Because we know that it won't work right anyway.

Good idea. Can I do this as part of the next PR?

PR #656, opened 2021-04-08
U318-015 Add "superproject" support.

godunko:


reznikmm (approved):

gnat/lsp.gpr
```diff
@@ -32,6 +32,9 @@ project LSP is
    );
    Build_Mode : Any_Build_Mode := external ("BUILD_MODE", "dev");
 
+   Superproject :=
+     external ("LSP_SUPERPROJECT", external ("SUPERPROJECT", ""));
```
reznikmm, line 36: I don't think we really need a batch of `XXX_SUPERPROJECT` (per project setting) currently.
I propose to remove them and keep just one `SUPERPROJECT`

reznikmm (changes requested):

reznikmm (approved):

PR #661, opened 2021-04-13
Update test baselines

setton:


reznikmm:
I think, we can drop these line altogether from tests.

setton:
> I think, we can drop these line altogether from tests.

I can replace thess with an <ANY> tag in the "documentation" field, shall I do this?

reznikmm (approved):

PR #665, opened 2021-04-14
U315-010 Add `shell` command to ALS tester driver

reznikmm:
Use this comman to create symbolic link during test run.
Fix regexp for `${NAME}` macro in tests. Add `DIR`
into environment variables.

### DIFF
--- a/source/tester/tester-tests.adb
+++ b/source/tester/tester-tests.adb
@@ -17,12 +17,17 @@
 
 with Ada.Command_Line;
 with Ada.Directories;
+with Ada.Streams;
 with Ada.Text_IO;
 with Ada.Strings.Unbounded; use Ada.Strings.Unbounded;
 with GNAT.OS_Lib;           use GNAT.OS_Lib;
 
 with GNATCOLL.Utils; use GNATCOLL.Utils;
 with GNATCOLL.JSON;  use GNATCOLL.JSON;
+
+with VSS.Stream_Element_Vectors;
+with VSS.Strings.Conversions;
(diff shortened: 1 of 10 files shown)


source/tester/tester-tests.adb
```diff
@@ -155,6 +159,96 @@ package body Tester.Tests is
...
+         Found : GNAT.OS_Lib.String_Access :=
+           GNAT.OS_Lib.Locate_Exec_On_Path (Name);
+      begin
+         return Result : constant String := Found.all do
+            Free (Found);
+         end return;
+      end To_Program;
+
+      List  : constant GNATCOLL.JSON.JSON_Array := Command.Get;
+      Cmd   : constant String := To_Program (GNATCOLL.JSON.Get (List, 1).Get);
```
setton, line 181: let's try running $SHELL if it exists with the given command line, rather than read a program from the list? if people want to run something more complex than a shell script, they can use a wrapper.

Alternatively we can decide to only support launching a Python script, perhaps more portable.

setton (changes requested):
change needed to support msys2 on the GHAs

source/tester/tester-tests.adb
reznikmm: We can easily run the Python (or any $SHELL) with this "shell" command. In the test I replaced the shell command with Python and added it to the README.md as example

        "shell": ["${PYTHON}", "${DIR}/makelink.py" ]

Is it good enough?

setton (approved):

PR #673, opened 2021-04-20
Replace BinTray with AWS S3

reznikmm:
Merge this after libadalang stable binaries uploaded to AWS S3.


setton (changes requested):
I propose to not drop gitpod support - let's just uncomment the lines that rely on bintray for now.

We can do two things to replace them in the future:
  - build the packages during gitpod initialisation instead of getting them from bintray
  - discuss with IT and find a way for the public to get readonly access to the s3 bucket

setton (approved):

PR #680, opened 2021-04-29
U429-030: Fix signatureHelp active parameter + Prefix notation

AdrienBoulanger:
Handle prefix notation: differentiate between a paremeter prefix
and a package prefix.
Properly select the active parameter + handle designators.

Add a test


AnthonyLeonardoGracio (changes requested):
We miss a test for this case

AnthonyLeonardoGracio (approved):

PR #706, opened 2021-05-24
U511-009: Fix refactoring import capatibility

AdrienBoulanger:
Properly check the client capabilities and fix the code handling for
non versioned files.

Use Get_Insert_With_Location.

Add a test.


source/ada/lsp-ada_handlers-named_parameters_commands.adb
```diff
@@ -107,21 +108,28 @@ package body LSP.Ada_Handlers.Named_Parameters_Commands is
...
          Edit.newText := Name & " => ";
 
-         if Client_Supports_documentChanges then
+         if Message_Handler.Versioned_Documents then
             Edits.documentChanges (1).Text_Document_Edit.edits.Append (Edit);
          else
-            Edits.changes (Edits.changes.First).Append
-              (LSP.Messages.TextEdit (Edit));
+            if Edits.changes.Contains (Self.Where.textDocument.uri) then
+               Edits.changes (Self.Where.textDocument.uri).Append
```
reznikmm, line 115: I think this will never be called because we already appended an empty vector at line:
```
Edits.changes.Insert (Document.URI, Empty);
```

source/ada/lsp-ada_handlers-refactor_imports_commands.adb
```diff
@@ -227,68 +233,67 @@ package body LSP.Ada_Handlers.Refactor_Imports_Commands is
...
-               end loop Already_Imported_Loop;
-
-               if not Already_Imported then
-
+            if S /= Libadalang.Slocs.No_Source_Location then
+               Edit.span := LSP.Lal_Utils.To_Span (S);
+               if Last then
                   Edit.newText := LSP.Types.To_LSP_String
-                    (ASCII.LF & LSP.Types.To_UTF_8_String
+                    ("" & ASCII.LF
```
reznikmm, line 271: it's not clear to me why do we need `"" &`? Looks like an extra do-nothing-operation.

PR #707, opened 2021-05-26
U526-026 Silence notification for imprecise navigation

### LINKED ISSUE #691: Constant "Imprecise Fallback" Popups When Editing
First of all: thanks for creating this extension - it's great to be able to work in Ada in such a familiar environment.

I'm using v22.0.7 of the extension in the version 1.55 of VS Code.

But, when editing code I get a constant stream of pop-ups with `Imprecise fallback used to compute cross-references on entity at:` messages.

Is it possible to suppress these pop-ups?

Thanks.

setton:
These messages are noisy and most often uninteresting.

Add a trace to allow printing them, and checking for them
in tests.

Closes #691

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -77,6 +77,12 @@ package body LSP.Ada_Handlers is
                              GNATCOLL.Traces.On);
    --  Trace to activate the support for incremental text changes.
 
+   Notifications_For_Imprecise : constant GNATCOLL.Traces.Trace_Handle :=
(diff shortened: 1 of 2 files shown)


reznikmm (approved):
Another solution could be a new configuration setting, so a user can turn it on/off by the vscode settings interface.

setton:
> Another solution could be a new configuration setting, so a user can turn it on/off by the vscode settings interface.

Good point. I thought about it, and think it's better not to clobber the documentation (and the vscode settings interface) for features that are purely intended for developers, so I'll push without. If we change our mind, it'll be easy to add later.

PR #721, opened 2021-06-09
U422-002 Defensive code against empty `Pretty_Printer` settings

ogorodnik:


source/ada/lsp-ada_contexts.adb
```diff
@@ -660,18 +664,40 @@ package body LSP.Ada_Contexts is
             Value            => Options,
             Is_Default_Value => Default);
 
-         --  Initialize an empty gnatpp command line object
+         --  Initialize an gnatpp command line object
+         Last := Options'First - 1;
```
reznikmm, line 668: I think we can reuse Options and drop Validated altogether like this:
```ada
for Item of Options.all loop
   if Item /= null and then Item.all /= "" then
      Last := Last + 1;
      Options (Last) := Item;
   end if;
end loop;
```

Then call `Parse` with `Options(Options'First .. Last)`.

reznikmm (changes requested):

reznikmm (approved):

PR #724, opened 2021-06-15
U614-038 Improve support for files detached from context

setton:
The modification for files are applied only in the contexts
where the files belong. When there is no project loaded and
the file is in a separate directory, no context accepts this
file, and the modifications are not taken into account.

Fix this by making the "fallback" context accept all
files.


petacreepers23:
When this will be available for download? would be very useful.

reznikmm (approved):

AnthonyLeonardoGracio:
@setton let's push this no?

PR #733, opened 2021-07-02
T714-004 Pass empty string to avoid precondition failure.

godunko:


reznikmm (approved):

godunko:
Not necessary, VSS has been fixed,

PR #734, opened 2021-07-05
topic/U705-017

joaopsazevedo:


reznikmm (approved):
conversion `function` to `const` looks strange, but we can work on it later

PR #752, opened 2021-07-29
U721-016: Duplicated profile in signatureHelp response

AdrienBoulanger:
Libadalang.Analysis.P_All_Env_Elements can return multiple times the
same node => do a filtering pass to remove duplicated node.

Also fix a wrong conversion when handling ErrorStmt.

### DIFF
--- a/source/ada/lsp-ada_contexts.adb
+++ b/source/ada/lsp-ada_contexts.adb
@@ -32,8 +32,6 @@ with LSP.Lal_Utils;               use LSP.Lal_Utils;
 with Libadalang.Common;           use Libadalang.Common;
 with Libadalang.Project_Provider;
 
-with Laltools.Common;
-
 with Langkit_Support.Slocs;
 
 with Utils.Command_Lines.Common;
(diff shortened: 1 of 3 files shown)


source/ada/lsp-ada_contexts.adb
```diff
@@ -467,14 +467,43 @@ package body LSP.Ada_Contexts is
    ---------------------------
 
    function Find_All_Env_Elements
-     (Self     : Context;
-      Name     : Libadalang.Analysis.Name;
-      Seq      : Boolean := True;
-      Seq_From : Libadalang.Analysis.Ada_Node'Class :=
-        Libadalang.Analysis.No_Ada_Node)
-     return Libadalang.Analysis.Ada_Node_Array is
```
AnthonyLeonardoGracio, line 475: Let's use sets instead

AnthonyLeonardoGracio (changes requested):

source/ada/lsp-ada_contexts.adb
AdrienBoulanger: This is now using a set for filtering duplicate and a vector to preserve the order.

AnthonyLeonardoGracio (approved):
Need to be pushed on edge though

PR #755, opened 2021-08-03
Don't send invisible symbols in completion after a dot

reznikmm:


godunko (approved):

AnthonyLeonardoGracio (changes requested):
Let's wait for feedback on U802-015 first.

Also, I don't see any ticket in the commit here (expected?)

reznikmm:
Hello @AnthonyLeonardoGracio,

> Let's wait for feedback on U802-015 first.

This is mostly unrelated to U802-015. Another completion function tries to find invisible symbols started with "." and always fails to find any. So it just spends CPU time.

> Also, I don't see any ticket in the commit here (expected?)

I've put U622-021 into the commit message.

AnthonyLeonardoGracio:
> > Let's wait for feedback on U802-015 first.
> 
> This is mostly unrelated to U802-015. Another completion function tries to find invisible symbols started with "." and always fails to find any. So it just spends CPU time.

Oh ok ;)

> > Also, I don't see any ticket in the commit here (expected?)
> 
> I've put U622-021 into the commit message.

Great, approved then.

AnthonyLeonardoGracio (approved):

PR #759, opened 2021-08-08
[VS Code] Fix snippet indentation

### LINKED ISSUE #694: Snippets do not honor my setting for indentation
When inserting snippets from this extension such as for, if, and so on, the placeholder blank lines where one would insert code are indented only three spaces where I have set my preferences for four spaces. Visual Studio Code 1.56.0.

Snippets from the Ada-snippets extension honor my four-space setting.

BTW, I wish Ada-snippets and Language Support for Ada (this one) would merge their snippet efforts. Both have their strengths but it is a nuisance to have to pick through both sets of options and pick the best one for each situation. Frankly, Ada-snippets seems a bit more robust.

Aerijo:
fixes #694

Replaces the hard coded triple-spaces in snippets with `\t`, which VS Code replaces with the configured indentation when the snippet is expanded.

Examples with `package` snippet (may need to copy somewhere that shows invisibles)
```
-- 3 spaces
package Name is
   pragma Preelaborate;
private
end Name;

-- 8 spaces
package Name is
        pragma Preelaborate;
private
end Name;

-- tabs
package Name is
	pragma Preelaborate;
private
end Name;
```

### DIFF
--- a/integration/vscode/ada/ada-snippets.json
+++ b/integration/vscode/ada/ada-snippets.json
@@ -4,7 +4,7 @@
         "scope": "ada",
         "body": [
             "accept ${1:Name} do",
-            "   $0",
+            "\t$0",
             "end ${1:Name};"
         ],
         "description": "Accept Statement"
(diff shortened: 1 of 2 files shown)


reznikmm (approved):
Excellent! :+1:

PR #768, opened 2021-08-17
topic/U415-027

joaopsazevedo:
When renaming a top level declaration both the spec and the body files will be renamed according to the standard naming conventions.


reznikmm (commented):
:+1:  This is a long awaited improvement! Great!

I guess, we should check `workspace.workspaceEdit.resourceOperations` client capability before sending rename file operations, because clients may not support them.

Also I propose to wait until stabilization phase ends before merging this PR.

reznikmm (approved):

PR #779, opened 2021-08-25
U415-027 Added support to resourceOperations client capability

joaopsazevedo:
File creations/deletions/renames are sent according to the capabilities
sent by the client in `workspace.workspaceEdit.resourceOperations`.
Added a test case where a top level declartion is renamed and file renames
are computed but not sent.


source/ada/lsp-lal_utils.adb
```diff
@@ -789,62 +811,104 @@ package body LSP.Lal_Utils is
...
+         end if;
 
          --  File deletions
 
-         if not Rename then
-            while Unbounded_String_Ordered_Sets.Has_Element
-              (File_Deletions_Cursor)
-            loop
+         if Versioned_Documents and then Is_Delete_Supported then
+            if not Rename then
```
reznikmm, line 858: I usually swap `then` and `else` branches for `if not <Condition>` to simplify reading.
See an example in [Ada_Style_Guide](https://en.wikibooks.org/wiki/Ada_Style_Guide/Programming_Practices#Statements) under *Nesting* header.

reznikmm (approved):

source/ada/lsp-lal_utils.adb
joaopsazevedo: I agree.
In this case, the only difference between the two branches is:
`uri    => File_URI`
or
`uri    => File_URI & ".bak"`
Therefore, I'll remove the if statement, all the duplicated code, and do:
`(if Rename then File_URI & ".bak" else File_URI)`

reznikmm (approved):

PR #785, opened 2021-08-31
U617-025: Index runtime files at startup

AnthonyLeonardoGracio:
To allow having completion for runtime packages that are not
included yet, which is similar to what the have with the old engine.


source/tester/tester-tests.adb
```diff
@@ -33,7 +33,7 @@ with Spawn.Processes.Monitor_Loop;
 
 package body Tester.Tests is
 
-   Max_Wait : constant := 4_000;
+   Max_Wait : constant := 40_000;
```
setton, line 36: I think this should be changed back to 4_000, in light of the new commit for the testsuite

testsuite/ada_lsp/completion.invisible.runtime/test.yaml
```diff
@@ -0,0 +1,3 @@
+title: 'completion.invisible.runtime'
+skip:
+    - ['SKIP', 'env.build.os.name != "linux"']
```
setton, line 3: could you add a comment? (should we skip this on Mac instead?)

setton (changes requested):

testsuite/ada_lsp/completion.invisible.runtime/test.yaml
AnthonyLeonardoGracio: I want to test on only one platform because runtime source locations may differ between each of them. I can add a comment saying that.

source/tester/tester-tests.adb
AnthonyLeonardoGracio: Right

setton (approved):

PR #786, opened 2021-08-31
U805-023: use previous context in signatureHelp

AdrienBoulanger:
If available try to reuse the signatureHelpContext which contains
the previous signatures: it's faster to just filter them that
recompute a new list of signatures.

Edge case for "\b": removing a character can make more signatures
valid and they could not be contained in the previous signature =>
always recompute the list.

Adapt tests: retriggerCharacters have changed
Add a new test.


source/ada/lsp-ada_handlers.adb
```diff
@@ -2494,11 +2494,33 @@ package body LSP.Ada_Handlers is
...
       procedure Add_Signature (Decl_Node : Libadalang.Analysis.Basic_Decl);
+      --  Add the signature of Decl_Node if it matches the current context.
+
+      procedure Filter_Signature
+        (Signature_Info : SignatureInformation;
+         Designator     : Libadalang.Analysis.Ada_Node;
+         Position       : LSP.Types.LSP_Number;
+         Is_Active      : Boolean);
+      --  Filter the signatures returned by the client using the current
+      --  current Position and designator.
```
AnthonyLeonardoGracio, line 2511: should be removed

AnthonyLeonardoGracio (changes requested):

source/ada/lsp-ada_handlers.adb
AdrienBoulanger: Remove the comment or replace "filter" by "remove"

PR #788, opened 2021-09-01
Revert "Remove test for navigation fallback on master"

setton:
This test should be applied on edge.

This reverts commit 29b3db4978cbe21956f86d1463d29f50cac64007.


reznikmm:
It looks like the tase waits for:
```
"type": 3,
"message": "Imprecise fallback used to compute cross-references on entity at:\n   main.adb\n   line: 18\n   column: 3"
```
but gets
```
"type": 2,
"message": "The result of 'definition' is approximate."
```

PR #799, opened 2021-09-08
U622-021: Fix bug in lazy completion and add a test

AnthonyLeonardoGracio:
. Some completion items (predefined keywords, attributes...) already
  have their doc pre-computed, so don't try to resolve it lazily.

. Add an automatic test.


AnthonyLeonardoGracio:
The regressions can be ignored, not due to my patch

AdrienBoulanger (approved):

PR #802, opened 2021-09-09
T826-026 Workspace_Symbols Request

ogorodnik:
Do not process too small pattern because this produces a huge
 response that is useless and costs a while.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -3744,6 +3744,9 @@ package body LSP.Ada_Handlers is
       Request : LSP.Messages.Server_Requests.Workspace_Symbols_Request)
       return LSP.Messages.Server_Responses.Symbol_Response
    is
+      use type LSP.Messages.Search_Kind;
+      use type VSS.Strings.Character_Count;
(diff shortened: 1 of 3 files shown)


doc/workspace_symbol_params.md
```diff
@@ -36,6 +36,7 @@ We extend the `WorkspaceSymbolParams` by adding extra fields:
   - whole_word to match the whole word instead of a part of it when True
   - negate to invert matching
   - kind is a search kind described above
+  - uri returns entities only for this file
```
reznikmm, line 39: In my opinion URI shouldn't be present in Workspace Symbol request. Use Document Symbol request instead. 

I approve this as a temporary solution, but if you merge the PR I'll open a TN to fix this latter.

reznikmm (dismissed):

doc/workspace_symbol_params.md
ogorodnik: I've already prepared a new patch without this "uri" and in the future Document Symbol request will be expanded to accept filtering options

reznikmm (approved):

PR #804, opened 2021-09-13
U910-043: Fix parsing error for optional client capabilities

AdrienBoulanger:
Some client capabilities like "moniker" were not defined as
optional => The server was crashing if their value was "null".


reznikmm (commented):
It would be nice to have a test

reznikmm (approved):

PR #815, opened 2021-09-21
U910-020 Partial response for Workspace_Symbols request

ogorodnik:


source/protocol/lsp-client_notification_receivers.ads
```diff
@@ -17,10 +17,13 @@
 --
 --  Interface to process notifications sent to the client.
 
+with LSP.Types;
 with LSP.Messages;
 
 package LSP.Client_Notification_Receivers is
 
+   type Progress_Value_Kind is (Params, SymbolInformation);
```
reznikmm, line 25: Can we use `Work_Done` instead of `Params`?

reznikmm (approved):

PR #816, opened 2021-09-22
U922-018: Don't send publishDiagnostics on didClose

AdrienBoulanger:
We were sending diagnostics after the document was already closed,
at this point this is the responsability of the client to free
the diagnostics (without receiving an empty list).

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -1976,7 +1976,6 @@ package body LSP.Ada_Handlers is
    is
       URI      : LSP.Messages.DocumentUri renames Value.textDocument.uri;
       File     : constant GNATCOLL.VFS.Virtual_File := Self.To_File (URI);
-      Diag     : LSP.Messages.PublishDiagnosticsParams;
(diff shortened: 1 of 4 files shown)


reznikmm (changes requested):
> at this point this is the responsability of the client

No, LSP doesn't work this way. According to specification:

> Diagnostics are “owned” by the server so it is the server’s responsibility to clear them if necessary. The following rule is used for VS Code servers that generate diagnostics:


> * if a language is single file only (for example HTML) then diagnostics are cleared by the server when the file is closed. Please note that open / close events don’t necessarily reflect what the user sees in the user interface. These events are ownership events. So with the current version of the specification it is possible that problems are not cleared although the file is not visible in the user interface since the client has not closed the file yet.
> * if a language has a project system (for example C#) diagnostics are not cleared when a file closes. When a project is opened all diagnostics for all files are recomputed (or read from a cache).

Currently we implement the first approach, because it simpler. If we want to switch to the second, we should
* clear diagnostics if project changed
* still clear diagnostics for an out-of-project file on closing it, because otherwise diagnostics stay forever on the client site.

PR #827, opened 2021-10-05
U922-009 Don't index files after `shutdown` request

reznikmm:
because it's useless and causes timeout on some tests.


AnthonyLeonardoGracio (changes requested):
Are you sure that it's the good TN?

AnthonyLeonardoGracio (approved):

PR #830, opened 2021-10-06
UA06-031: Don't check for tokens in Is_End_Label filter

AnthonyLeonardoGracio:
Since we are only trying to detect end name nodes.


reznikmm (changes requested):
consider this code:
```ada
package Bar is
end |;
```

Here we don't have End_Name node in the tree at all, so we need to check previous token to detect, that we complete end name.

PR #853, opened 2021-11-11
UB11-011 Call "npm audit" to check for vulnerabilities

setton:
And interrupt the build if any is found.

### DIFF
--- a/.github/workflows/build-binaries.sh
+++ b/.github/workflows/build-binaries.sh
@@ -32,6 +32,13 @@ if [[ ${GITHUB_REF##*/} != 2*.[0-9]*.[0-9]* ]]; then
     git rebase --verbose origin/edge
 fi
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-binaries.sh
```diff
@@ -32,6 +32,13 @@ if [[ ${GITHUB_REF##*/} != 2*.[0-9]*.[0-9]* ]]; then
     git rebase --verbose origin/edge
 fi
 
+# Audit the npm packages
+cd integration/vscode/ada
+npm install
+# Run npm audit to check for any vulnerabilities - bail out if this returns nonzero
+npm audit || exit 1
```
reznikmm, line 39: Let's remove `|| exit 1`, because we have `set  -e` in this script.

reznikmm (dismissed):

PR #856, opened 2021-11-25
UA28-007 Added file operation requests and notifications

joaopsazevedo:
Added different behaviours depending on the change kind.
Improved DidChangeWatchedFiles Notification.
Added test cases.


source/ada/lsp-ada_handlers.ads
```diff
@@ -255,6 +273,8 @@ private
       Completion_Snippets_Enabled : Boolean := False;
       --  True if the client supports completion snippets
 
+      Client_Capabilities : LSP.Messages.ClientCapabilities;
```
reznikmm, line 276: We already keep `ClientCapabilities` in `Message_Handler.Client.capabilities`. Let's replace `Client_Capabilities` with a function, that returns `Self.Client.capabilities`?

source/ada/lsp-ada_handlers.ads
```diff
@@ -309,14 +329,15 @@ private
       File_Monitor    : LSP.File_Monitors.File_Monitor_Access;
       --  Filesystem monitoring
 
+      Dynamic_Registration_Ids : LSP.Ada_Handlers.Dynamic_Registration_Ids;
```
reznikmm, line 332: A short comment is appreciated here. Also, I'm not sure if it worse a dedicated field at all, because it's used only once.

reznikmm (changes requested):

source/ada/lsp-ada_handlers.ads
joaopsazevedo: That's right, I didn't note that there was already a `Message_Handler.Client`. So yes,
I'll add this function.

source/ada/lsp-ada_handlers.ads
joaopsazevedo: Agree.
I'll remove this and hard code the ids when we register the capabilities.
If we ever need to deregister capabilities, then we come back to this approach to avoid duplicated hard coded strings.

joaopsazevedo: Done

reznikmm (approved):

PR #909, opened 2022-02-07
topic/SA11-047 Extract Subprogram initial implementation

joaopsazevedo:


source/ada/lsp-ada_handlers.adb
```diff
@@ -1351,6 +1353,61 @@ package body LSP.Ada_Handlers is
 
          --  Refactoring Code Actions
 
+         --  Extract Subprogram
+         declare
```
setton, line 1357: this begin block is big enough: worth extracting to a local subprogram!! (using this feature!)

setton (changes requested):

setton (approved):

PR #910, opened 2022-02-17
V216-030: Fix attribute completion in declarations

AnthonyLeonardoGracio:
We should not rely on the current node's text to be the completion
prefix when the user just typed "'". Instead, we should set it to
null in order to list all the possible attributes.

Add an automatic test.

### DIFF
--- a/testsuite/ada_lsp/V216-030.completion.attribute_in_decl/test.json
+++ b/testsuite/ada_lsp/V216-030.completion.attribute_in_decl/test.json
@@ -1,0 +1,286 @@
+[
+   {
+      "comment": [
+         "test automatically generated"
+      ]
+   },
+   {
+      "start": {
+         "cmd": [
+            "${ALS}"
+         ]
+      }
+   },
+   {
+      "send": {
+         "request": {
+            "jsonrpc": "2.0",
+            "id": 1,
+            "method": "initialize",
+            "params": {
+               "processId": 30104,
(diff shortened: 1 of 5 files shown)


testsuite/ada_lsp/V216-030.completion.attribute_in_decl/test.json
```diff
@@ -0,0 +1,292 @@
...
+                     "typeDefinitionProvider": true,
+                     "implementationProvider": true,
+                     "referencesProvider": true,
+                     "documentHighlightProvider": true,
+                     "documentSymbolProvider": true,
+                     "codeActionProvider": {},
+                     "documentFormattingProvider": true,
+                     "renameProvider": {},
+                     "foldingRangeProvider": true,
+                     "executeCommandProvider": {
```
joaopsazevedo, line 112: I would remove the `executeCommandProvider` because it does not affect this test and when we add a new command we will have the adjust this.

joaopsazevedo (approved):
Just a minor observation on the `executeCommandProvider` field of the test.
Can be done in another commit, so feel free to merge.

testsuite/ada_lsp/V216-030.completion.attribute_in_decl/test.json
AnthonyLeonardoGracio: Ok I will change that

joaopsazevedo (approved):

PR #920, opened 2022-03-15
V303-003 Do not index non-Ada files, for memory reasons

setton:


source/ada/lsp-ada_contexts.adb
```diff
@@ -858,17 +903,19 @@ package body LSP.Ada_Contexts is
...
       PLE     : Boolean := True)
    is
-      Unit : constant Libadalang.Analysis.Analysis_Unit :=
-        Self.LAL_Context.Get_From_File
-          (File.Display_Full_Name,
-           Charset => Self.Get_Charset,
-           Reparse => Reparse);
+      Unit : Libadalang.Analysis.Analysis_Unit;
    begin
       Trace
```
AnthonyLeonardoGracio, line 908: Let's put the trace under the early return, when we are sure that we are actually indexing it

AnthonyLeonardoGracio (changes requested):

source/ada/lsp-ada_contexts.adb
setton: Good point, I added a trace when we're bailing in the early return.

PR #946, opened 2022-04-21
Topic/v419 041.prepare call hierarchy.outgoing calls

AdrienBoulanger:
From the LSP documentation:
	/**
	 * The range at which this item is called. This is the range relative to
	 * the caller, e.g the item passed to `callHierarchy/outgoingCalls` request.
	 */
This is problematic for Ada where the spec and body can be in two different files: use prepareCallHierarchy request to return an item corresponding to the body where the outgoingCalls are located.


reznikmm:
All calls are located in the caller's body. We return location of the caller's body and so we meet API.

reznikmm (approved):

PR #951, opened 2022-04-21
UB26-005 Add task for proof of subprogram

yannickmoy:
Use promises as getting all symbols returns a promise.


reznikmm (approved):
Superb!

PR #976, opened 2022-06-09
V609-058: Add useCompletionSnippets flag

AnthonyLeonardoGracio:
And an associated VS Code preference, disabled by default.

### DIFF
--- a/source/ada/lsp-ada_handlers.ads
+++ b/source/ada/lsp-ada_handlers.ads
@@ -268,6 +268,10 @@ private
       Completion_Snippets_Enabled : Boolean := False;
       --  True if the client supports completion snippets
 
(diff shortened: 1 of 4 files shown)


source/ada/lsp-ada_handlers.ads
```diff
@@ -268,6 +268,10 @@ private
       Completion_Snippets_Enabled : Boolean := False;
       --  True if the client supports completion snippets
 
+      Use_Completion_Snippets : Boolean := True;
```
setton, line 271: should we set this to False by default? so that the testsuite tests the default IDE behavior? (unless tests specifically activate snippets)

setton (changes requested):
looks good, just one meta question

setton (approved):

PR #982, opened 2022-06-16
Add partial gnatpp option to On_Range_Formatting_Request handler

joaopsazevedo:
When setting the ALS.PARTIAL_GNATPP=yes trace, On_Range_Formatting_Request now uses partial gnatpp to perform a range formatting.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -119,7 +119,7 @@ package body LSP.Ada_Handlers is
 
    Partial_GNATpp : constant GNATCOLL.Traces.Trace_Handle :=
      GNATCOLL.Traces.Create ("ALS.PARTIAL_GNATPP",
-                             GNATCOLL.Traces.Off);
+                             GNATCOLL.Traces.On);
    --  Use partial formatting mode of gnatpp if On. Otherwise, use diff
    --  algorithm.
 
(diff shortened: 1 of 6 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -117,6 +117,12 @@ package body LSP.Ada_Handlers is
                              GNATCOLL.Traces.On);
    --  Trace to enable/disable runtime indexing. Useful for the testsuite.
 
+   Partial_GNATpp : constant GNATCOLL.Traces.Trace_Handle :=
+     GNATCOLL.Traces.Create ("ALS.PARTIAL_GNATPP",
+                             GNATCOLL.Traces.Off);
```
reznikmm, line 122: I would use it `On` by default. Currently, GNAT Studio doesn't use this request any way.

reznikmm: Can we merge this into `master` instead of `edge`, BTW?

joaopsazevedo: > I would use it `On` by default. Currently, GNAT Studio doesn't use this request any way.

> @reznikmm I just checked and actually GS tries to send the request, if the server supports it. I think we can enable it by default on the server side, and add a specific trace to disable it on GS side for now though.

I now set `ALS.PARTIAL_GNATPP` to `On` by default which will set `documentFormattingProvider` to `True`.
One test baseline has to be updated and note that partial gnatpp is formatting more lines that those requested. This was already discussed last week and we're implementing a solution soon.

> Can we merge this into `master` instead of `edge`, BTW?

This one I don't think we can since this requires the latest laltools in our internal nightly tests.

PR #983, opened 2022-06-17
V614-021: Use completion weight in sortText

AnthonyLeonardoGracio:
In order to place items with an higher weight at the
top of the completion window.


reznikmm (commented):
I'm afraid items with weight `99` will win items with weight `100` :(

AnthonyLeonardoGracio:
I confirm, padding is added. Submitting.

PR #996, opened 2022-07-07
V704-034: Limit number of logs in $ALS_HOME/.als

AdrienBoulanger:
Add logThreshold property to limit the number of logs.

### DIFF
--- a/source/ada/lsp-ada_handlers.adb
+++ b/source/ada/lsp-ada_handlers.adb
@@ -531,6 +531,30 @@ package body LSP.Ada_Handlers is
       LSP.File_Monitors.Unchecked_Free (Self.File_Monitor);
    end Cleanup;
 
+   ----------------
+   -- Clean_Logs --
+   ----------------
+
+   procedure Clean_Logs (Self : access Message_Handler; Dir : Virtual_File) is
+      Files : File_Array_Access := Read_Dir (Dir, Files_Only);
+      Dummy : Boolean;
+      Cpt   : Integer := 0;
+   begin
+      Sort (Files.all);
+      --  Browse the log files in reverse timestamp order
+      for F of reverse Files.all loop
+         --  Filter out files like traces.cfg
+         if GNATCOLL.Utils.Ends_With (+F.Base_Name, ".log") then
+            Cpt := Cpt + 1;
+            --  Delete the old logs
+            if Cpt > Self.Log_Threshold then
+               Delete (F, Dummy);
+            end if;
+         end if;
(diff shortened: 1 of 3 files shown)


source/ada/lsp-ada_handlers.adb
```diff
@@ -531,6 +531,30 @@ package body LSP.Ada_Handlers is
...
+   procedure Clean_Logs (Self : access Message_Handler; Dir : Virtual_File) is
+      Log_Files : File_Array_Access := Read_Dir (Dir, Files_Only);
+      Dummy     : Boolean;
+      Cpt       : Integer := 0;
+   begin
+      Sort (Log_Files.all);
+      --  Browse the log files in reverse timestamp order
+      for F of reverse Log_Files.all loop
+         --  Filter out files like traces.cfg
+         if GNATCOLL.Utils.Starts_With (+F.Base_Name, "als.") then
```
AnthonyLeonardoGracio, line 547: You should filter with .txt instead: the log filenames may be changed in traces.cfg

source/ada/lsp-ada_handlers.adb
```diff
@@ -531,6 +531,30 @@ package body LSP.Ada_Handlers is
       LSP.File_Monitors.Unchecked_Free (Self.File_Monitor);
    end Cleanup;
 
+   ----------------
+   -- Clean_Logs --
+   ----------------
+
+   procedure Clean_Logs (Self : access Message_Handler; Dir : Virtual_File) is
+      Log_Files : File_Array_Access := Read_Dir (Dir, Files_Only);
```
AnthonyLeonardoGracio, line 539: "Files" might be a better name, since it may not only contain log files

AnthonyLeonardoGracio (changes requested):

source/ada/lsp-ada_handlers.adb
AdrienBoulanger: The default name is finishing by ".log" and we are overriding the name in traces.cfg when using ALS_HOME

source/ada/lsp-ada_handlers.adb
AdrienBoulanger: Done

source/ada/lsp-ada_handlers.adb
AnthonyLeonardoGracio: Then check for '.log' no ? Also, will it conflict with the GS preference?

AnthonyLeonardoGracio (approved):

PR #1001, opened 2022-07-14
UB30-006 Fix exception raised when range formatting the whole document

joaopsazevedo:
If the user selects from the procedure name `Main` to the procedure end label `Main`, the whole file needs to be formatted. This PR fixes an exception that was being raised when converting from `Source_Location_Range` to `Span`.
```ada
procedure Main is

begin
   declare
   A : Integer;
      begin
      null;
      end;

end Main;
```

### DIFF
--- a/source/ada/lsp-ada_documents.adb
+++ b/source/ada/lsp-ada_documents.adb
@@ -2456,18 +2456,26 @@ package body LSP.Ada_Documents is
       Start_Line      : constant LSP.Types.Line_Number :=
         LSP.Types.Line_Number (Segment.Start_Line) - 1;
       Start_Line_Text : constant VSS.Strings.Virtual_String :=
-        Self.Text.Slice
-          (Self.Line_To_Marker (Start_Line),
-           Self.Line_To_Marker (Start_Line + 1));
+        (if Self.Line_To_Marker.Last_Index = Start_Line then
+           Self.Text.Slice
+             (Self.Line_To_Marker (Start_Line), Self.Text.After_Last_Character)
+         else
+           Self.Text.Slice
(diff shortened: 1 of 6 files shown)


source/ada/lsp-ada_documents.adb
```diff
@@ -2456,18 +2456,26 @@ package body LSP.Ada_Documents is
       Start_Line      : constant LSP.Types.Line_Number :=
         LSP.Types.Line_Number (Segment.Start_Line) - 1;
       Start_Line_Text : constant VSS.Strings.Virtual_String :=
-        Self.Text.Slice
-          (Self.Line_To_Marker (Start_Line),
-           Self.Line_To_Marker (Start_Line + 1));
+        (if Self.Line_To_Marker.Last_Index = Start_Line then
+           Self.Text.Slice
+             (Self.Line_To_Marker (Start_Line), Self.Text.After_Last_Character)
```
godunko, line 2461: It is better to use Self.Text.At_Last_Character here

joaopsazevedo: That was my first try, however, I noticed that it misses the last character.
I just added a commit with the test case (I forgot to add it previously):
https://github.com/AdaCore/ada_language_server/commit/434fc5cd224776f58e5b893091a01c35478125b1#diff-984e603b643daea8b21d34cc3034804b40d57a78698060974d449fea8d495176R10
In this test case, note that there is no line break at the end of the file. Using `At_Last_Character` will result in a `;;` on the last line.
This is because the To_LSP_Range will be 1:1 to 10:9 instead of 10:10.

godunko (approved):

PR #1003, opened 2022-07-20
V628-024 (1) Symbol diff has been added

ogorodnik:


testsuite/ada_lsp/V628-024.symbol_diff/test.json
```diff
@@ -0,0 +1,275 @@
...
+                        "start": {
+                           "line": 9,
+                           "character": 0
+                        },
+                        "end": {
+                           "line": 9,
+                           "character": 0
+                        }
+                     },
+                     "newText": "\u0000"
```
reznikmm, line 214: I wouldn't join this patch before fixing insertion of binary zero into the text. In my opinion, it could break editors including GNAT Studio.

ogorodnik: I've already tested with GNAT Studio and there are not any issues with this.

AnthonyLeonardoGracio: Where this comes from?

ogorodnik: From gnatpp

AnthonyLeonardoGracio: The test fails also, needs to checked. @CKMonika is the "\u0000" that comes from gnatpp expected?

AnthonyLeonardoGracio: It's strange, I don't see any reason for that... Maybe @reznikmm has an idea?

AnthonyLeonardoGracio: Actually I think the test is failing due to a timing issue: the test seems to be stopped while indexing the project. You are probably missing a wait somewhere

reznikmm: > It's strange, I don't see any reason for that... Maybe @reznikmm has an idea?

because Andry tries to modify `master` while `edge` has its-own modification for range formatting and doesn't use diff any more:

```ada
   Partial_GNATpp : constant GNATCOLL.Traces.Trace_Handle :=
     GNATCOLL.Traces.Create ("ALS.PARTIAL_GNATPP",
                             GNATCOLL.Traces.On);
   --  Use partial formatting mode of gnatpp if On. Otherwise, use diff
   --  algorithm.
```

PR #1004, opened 2022-07-21
V603-024: Add two spaces for Ada comments in VS Code

AnthonyLeonardoGracio:


reznikmm (approved):
This approach works in most cases, except when you already have code like:
```ada
--comment with no space
```

Then it doesn't delete double-dash, but appends another `-- ` before the first double-dash.

In my opinion, this is still better then what we have before, however.

AnthonyLeonardoGracio:
@reznikmm I see. I think we had the same issue in GS for years, so yes, I think it's still better with this patch, until we find a more optimal solution.

PR #1013, opened 2022-08-02
V511-027: Run PP on completion snippet

AdrienBoulanger:
Encode snippet in pseudo Ada code, run PP and decode it.
Enforce PP switches to have the parameter aligned: it can
be disabled via a trace which is used to preserve all the tests
checking the snippets' contents.
Add a test with the trace disabled to check the PP behavior
on snippets.


AnthonyLeonardoGracio (approved):

AdrienBoulanger:
Move the patch to master

PR #1018, opened 2022-08-03
V628 024 2

ogorodnik:


source/ada/lsp-ada_documents.adb
```diff
@@ -718,19 +1035,39 @@ package body LSP.Ada_Documents is
...
+         else
+            declare
+               Formatted : constant VSS.Strings.Virtual_String :=
+                 VSS.Strings.Conversions.To_Virtual_String (S.all);
+               Slice     : VSS.Strings.Virtual_String;
+
+            begin
+               LSP.Lal_Utils.Span_To_Slice (Formatted, Out_Span, Slice);
+
+               Diff_Symbols
```
AnthonyLeonardoGracio, line 1056: Could we also use it in the Range_Formatting action (where it makes even more sense)?

AnthonyLeonardoGracio (changes requested):

AnthonyLeonardoGracio (approved):

PR #1021, opened 2022-08-05
V315-025: Convert to UTF-8 before preprocessing

AnthonyLeonardoGracio:
This is needed to handle other charsets (e.g: cyrillic).

Remove the old ad-hoc preprocessor implementation.


AnthonyLeonardoGracio:
It fixes the nightly regression on edge for S523-033:lsp:cyrillic

source/ada/lsp-ada_handlers-file_readers.adb
```diff
@@ -27,6 +32,104 @@ package body LSP.Ada_Handlers.File_Readers is
...
+               Output_Index := 1;
+               Iconv (State        => State,
+                      Inbuf        => Raw.all,
+                      Input_Index  => Input_Index,
+                      Outbuf       => Outbuf,
+                      Output_Index => Output_Index,
+                      Result       => Conv_Result);
+
+               --  Append the converted contents
+               if Decoded /= null then
```
godunko, line 81: Decoded is always null here.

I recommend to use VSS.Strings.Converters here instead of Iconv. We can add necessary encodings to VSS, it is not complicated; however, VSS's decoders is able to recover after conversion error, which may be important for us - it save as much information as possible, instead of dropping everything after first failure.

AnthonyLeonardoGracio: Indeed, It's Raw that I need to free, thanks.

For the VSS stuff, if you know how to proceed feel free to make the modifications! I can create a separate ticket for that. I just want to fix the testsuite meanwhile.

godunko (approved):

PR #1023, opened 2022-08-09
V809-013: Add 'subprogram box' command

AnthonyLeonardoGracio:
The command is based on the LSP DocumentSymbols request to
retrieve the enclosing subprogram, if any.

### DIFF
--- a/integration/vscode/ada/package.json
+++ b/integration/vscode/ada/package.json
@@ -418,6 +418,10 @@
             {
                 "command": "ada.otherFile",
                 "title": "Go to other Ada file"
+            },
+            {
+                "command": "ada.subprogramBox",
+                "title": "Add subprogram box"
             }
         ],
         "keybindings": [
--- a/integration/vscode/ada/src/extension.ts
+++ b/integration/vscode/ada/src/extension.ts
@@ -27,6 +27,7 @@ import * as process from 'process';
 import GPRTaskProvider from './gprTaskProvider';
 import cleanTaskProvider from './cleanTaskProvider';
 import gnatproveTaskProvider from './gnatproveTaskProvider';
+import { getSubprogramSymbol } from './gnatproveTaskProvider';
 import { alsCommandExecutor } from './alsExecuteCommand';
 import { ALSClientFeatures } from './alsClientFeatures';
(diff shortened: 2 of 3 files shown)


integration/vscode/ada/package.json
```diff
@@ -418,6 +418,10 @@
             {
                 "command": "ada.otherFile",
                 "title": "Go to other Ada file"
+            },
+            {
+                "command": "ada.subprogramBox",
+                "title": "Add subprogram box"
```
setton, line 424: let's call this "Add Ada subprogram box" ?

reznikmm (commented):
If we move the command logic to ALS then GS (and perhaps other editors) could reuse it, I guess.

integration/vscode/ada/src/extension.ts
```diff
@@ -41,6 +42,45 @@ let alsTaskProvider: vscode.Disposable[] = [
...
+        .then(symbol => {
+            if (symbol !== null) {
+                const name: string = symbol.name ?? ""
+                const insertPos = new vscode.Position(symbol.range.start.line, 0);
+                const indentationRange = new vscode.Range(insertPos, symbol.range.start)
+                const indentation: string = activeEditor?.document.getText(indentationRange) ?? ""
+
+                // Generate the subprogram box after retrieving the indentation of the line of
+                // the subprogram's body declaration.
+                let text: string = indentation + "---" + '-'.repeat(name.length) + "---\n"
```
setton, line 69: instead of using "\n" systematically, we should probably query the document for the right end of line character -- looks like there is something here https://code.visualstudio.com/api/references/vscode-api  (search for "eol")

setton (changes requested):
can't wait to use this! two questions

integration/vscode/ada/package.json
AnthonyLeonardoGracio: Ok

integration/vscode/ada/src/extension.ts
AnthonyLeonardoGracio: Indeed, I will have a look

setton (approved):

setton (approved):

PR #1024, opened 2022-08-10
V809-025: Make our commands Ada-specific

AnthonyLeonardoGracio:
By prefixing them by 'Ada: ' (like other extensions do) and
by adding when clauses (i.e: filters) for their associated key
shortcuts and the command palette.

### DIFF
--- a/integration/vscode/ada/package.json
+++ b/integration/vscode/ada/package.json
@@ -417,20 +417,37 @@
         "commands": [
             {
                 "command": "ada.otherFile",
-                "title": "Go to other Ada file"
+                "title": "Ada: Go to other file"
             },
             {
                 "command": "ada.subprogramBox",
-                "title": "Add subprogram box"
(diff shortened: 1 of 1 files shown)


integration/vscode/ada/package.json
```diff
@@ -417,20 +417,37 @@
         "commands": [
             {
                 "command": "ada.otherFile",
-                "title": "Go to other Ada file"
+                "title": "Ada: Go to other Ada file"
```
setton, line 420: let's call this one "Ada: Go to other file" ?

setton (dismissed):

integration/vscode/ada/package.json
AnthonyLeonardoGracio: Ah right, good catch

joaopsazevedo (approved):

joaopsazevedo (approved):

PR #1028, opened 2022-08-16
Topic/v808 019

AnthonyLeonardoGracio:


AnthonyLeonardoGracio:
Merges of edge -> master

joaopsazevedo (approved):
Since the last bump it matches correctly with `git log origin/master..origin/edge --oneline --no-merges` so it looks good to me.

PR #1030, opened 2022-08-16
V805-014-1 Fixing NULL in symbols diff

ogorodnik:


joaopsazevedo (dismissed):

reznikmm (changes requested):
Why not merge this into master? Stabilization is just finished, `master` = `edge` right?

PR #1036, opened 2022-08-22
V511-027: Run PP on completion snippet

AdrienBoulanger:
Encode snippet in pseudo Ada code, run PP and decode it.
When successfull, it will generate a textEdit for the
completionItem.

Enforce PP switches to have the parameter aligned: it can
be disabled via a trace which is used to avoid updating
all the snippet tests.

Add tests with the trace disabled to check the PP behavior
on snippets.

### DIFF
--- a/source/ada/lsp-ada_completions-generic_assoc.adb
+++ b/source/ada/lsp-ada_completions-generic_assoc.adb
@@ -37,7 +37,7 @@ package body LSP.Ada_Completions.Generic_Assoc is
       Names        : in out Ada_Completions.Completion_Maps.Map;
       Unsorted_Res : in out LSP.Messages.CompletionItem_Vector)
    is
-      pragma Unreferenced (Filter, Sloc, Names);
+      pragma Unreferenced (Filter, Names);
       use Libadalang.Analysis;
       use Libadalang.Common;
       use LSP.Ada_Completions.Generic_Assoc_Utils;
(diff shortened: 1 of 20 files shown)


source/ada/lsp-ada_completions-generic_assoc.adb
```diff
@@ -49,6 +49,11 @@ package body LSP.Ada_Completions.Generic_Assoc is
 
       Designators       : Laltools.Common.Node_Vectors.Vector;
 
+      Prefix      : VSS.Strings.Virtual_String;
+      --  The whole string before the snippet (including whitespaces)
```
AnthonyLeonardoGracio, line 53: Add new line

source/ada/lsp-ada_completions.adb
```diff
@@ -107,6 +138,308 @@ package body LSP.Ada_Completions is
...
+                   (Context  => Tmp_Context,
+                    Filename => "",
+                    Buffer   => Full,
+                    Rule     => Rule);
+               --  -2 because:
+               --  - align using the column before the cursor
+               --  - and "column = offset - 1"
+               Pp.Actions.Set_Partial_Gnatpp_Offset
+                 (Integer (Span.first.character) - 2);
+               Pp.Actions.Format_Vector
```
AnthonyLeonardoGracio, line 404: I think there is a new API for partial formatting, that simplifies things. See https://git.adacore-it.com/c/libadalang-tools/+/119537/2/src/laltools-partial_gnatpp.ads

Let's try to use it id possible

source/ada/lsp-ada_documents.adb
```diff
@@ -2541,6 +2544,23 @@ package body LSP.Ada_Documents is
             Insert_Text.Append (")$0");
 
             Item.insertText := (True, Insert_Text);
+            --  For performance reason, don't do a call to GNATpp for each
```
AnthonyLeonardoGracio, line 2547: Should we remove that commented code?

AnthonyLeonardoGracio (changes requested):

AnthonyLeonardoGracio (approved):

PR #1059, opened 2022-10-10
V725-006 Update formatted text last character offset

joaopsazevedo:
Companion patch for a change in partial gnatpp that no longer adds a line break to the end of the formatted text.


CKMonika (approved):
Looks good to follow up gnatpp partial formatting changed

PR #1061, opened 2022-10-11
VA05-011: Make lsp_client resilient to invalid json

AdrienBoulanger:
It was detected when GNATpp was sending message in stdout polluting the json stream and making the raw_client do an infinite loop (it was never consuming the "invalid json")
=> implement a basic defensive mechanism consuming the lines one by one until finding a proper json header

### DIFF
--- a/source/client/lsp-raw_clients.adb
+++ b/source/client/lsp-raw_clients.adb
@@ -29,6 +29,9 @@ with Spawn.Processes; use Spawn.Processes;
 
 package body LSP.Raw_Clients is
 
+   Parse_Exception : exception;
+   --   Local exception raised when parsing invalid data
+
    New_Line : constant String :=
      (Ada.Characters.Latin_1.CR, Ada.Characters.Latin_1.LF);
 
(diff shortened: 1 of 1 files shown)


source/client/lsp-raw_clients.adb
```diff
@@ -325,12 +325,24 @@ package body LSP.Raw_Clients is
...
-                     Delete
-                       (Client.Buffer, 1, Start + 2 * New_Line'Length - 1);
+                     begin
+                        Parse_Headers
+                          (Slice (Client.Buffer, 1, Start + 1),
+                           Client.To_Read);
+                        Delete
+                          (Client.Buffer, 1, Start + 2 * New_Line'Length - 1);
+                     exception
+                        when E : others =>
```
AnthonyLeonardoGracio, line 335: Can't we be more precise regarding the exception we expect?

AdrienBoulanger: Parse_Headers is raising Constraint_Error for now, I can add a custom Exception if needed

AnthonyLeonardoGracio (approved):

PR #1082, opened 2022-11-25
VB24-035: Add exception handler when computing doc

AnthonyLeonardoGracio:
Without logging the exception, which is too noisy. This code will be replaced by gnatdoc at some point.


godunko (changes requested):
There two more uses of Get_Documentation in lsp-ada_handlers.adb in On_Hower_Requesr and On_Signature_Help requests what need to be protected too.

godunko (approved):

PR #1084, opened 2022-11-27
Miscellaneous convenient changes.

Blady-Com:
Add convenient makefile targets in order to build and install glib lsp_client.
Ensure that address randomization is not active for backtracking in dev mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,3 +146,13 @@ check: all
 
 deploy: check
 	integration/$(USER)/deploy.sh $(PLATFORM)
+
+lsp_client_glib-build:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -146,3 +146,13 @@ check: all
 
 deploy: check
 	integration/$(USER)/deploy.sh $(PLATFORM)
+
+lsp_client_glib-build:
+	$(GPRBUILD) -P gnat/lsp_client_glib.gpr -p $(COVERAGE_BUILD_FLAGS) \
+		-XVERSION=$(TRAVIS_TAG)
```
reznikmm, line 152: We don't use travis, why we need `TRAVIS_TAG`?

PR #1089, opened 2022-12-09
UC15-036 Replace GNATCOLL.Projects with libgpr

fipgil:
UC15-036  Let ada_language_server use GPR2 instead of GNATCOLL.Projects.


reznikmm (approved):

reznikmm:
if this works in `master` I would suggest to push it into `master` (not `edge`). Philippe is checking this now.

fipgil:
This patch is not compiling on master because in requires let Safe_Renamer handle GNATCOLL & GPR2 projects. libadalang-tools patch. This patch is not available in stable-libadalang-tools used currently by als
At run time it requires modified gpr support reload capability not currently available in stable-gpr2.

PR #1090, opened 2022-12-12
VC09-024 Limit number of contexts where file is indexed.

godunko:


AnthonyLeonardoGracio (commented):
I think we need to find a better way to handle this: if the user has 400 different contexts, limiting to 10 will not help a lot, we'll index each time he opens a file from a different context..
I think we need to talk with the LAL team to see what we can do here.

PR #1106, opened 2023-01-04
VC16-002: Add use-clause completion provider

AnthonyLeonardoGracio:
This provider proposes the use-clause corresponding to the with-clause on the same line, if any.

### DIFF
--- a/source/ada/lsp-ada_completions-use_clauses.adb
+++ b/source/ada/lsp-ada_completions-use_clauses.adb
@@ -1,0 +1,125 @@
+------------------------------------------------------------------------------
+--                         Language Server Protocol                         --
+--                                                                          --
+--                     Copyright (C) 2018-2023, AdaCore                     --
+--                                                                          --
+-- This is free software;  you can redistribute it  and/or modify it  under --
+-- terms of the  GNU General Public License as published  by the Free Soft- --
+-- ware  Foundation;  either version 3,  or (at your option) any later ver- --
+-- sion.  This software is distributed in the hope  that it will be useful, --
+-- but WITHOUT ANY WARRANTY;  without even the implied warranty of MERCHAN- --
+-- TABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public --
+-- License for  more details.  You should have  received  a copy of the GNU --
+-- General  Public  License  distributed  with  this  software;   see  file --
+-- COPYING3.  If not, go to http://www.gnu.org/licenses for a complete copy --
+-- of the license.                                                          --
+------------------------------------------------------------------------------
+
+with VSS.Strings.Conversions;
+with Libadalang.Iterators;
+with LSP.Ada_Completions.Filters;
+
+package body LSP.Ada_Completions.Use_Clauses is
+
+   ------------------------
+   -- Propose_Completion --
+   ------------------------
+
+   overriding procedure Propose_Completion
+     (Self   : Use_Clause_Completion_Provider;
+      Sloc   : Langkit_Support.Slocs.Source_Location;
+      Token  : Libadalang.Common.Token_Reference;
+      Node   : Libadalang.Analysis.Ada_Node;
+      Filter : in out LSP.Ada_Completions.Filters.Filter;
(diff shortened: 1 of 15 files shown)


source/ada/lsp-ada_completions-use_clauses.ads
```diff
@@ -0,0 +1,35 @@
+------------------------------------------------------------------------------
+--                         Language Server Protocol                         --
+--                                                                          --
+--                     Copyright (C) 2018-2021, AdaCore                     --
```
AdrienBoulanger, line 4: 2018-2021 => 2023

source/ada/lsp-ada_completions-use_clauses.adb
```diff
@@ -0,0 +1,125 @@
+------------------------------------------------------------------------------
+--                         Language Server Protocol                         --
+--                                                                          --
+--                     Copyright (C) 2018-2021, AdaCore                     --
```
AdrienBoulanger, line 4: 2018-2021 => 2023

source/ada/lsp-ada_documents.adb
```diff
@@ -2837,8 +2838,8 @@ package body LSP.Ada_Documents is
 
       Filter : LSP.Ada_Completions.Filters.Filter;
    begin
-      if Parent.Is_Null
-        or else (Parent.Kind not in
+      if not Parent.Is_Null
```
AdrienBoulanger, line 2841: you removed the case where we do nothing when Parent.Is_Null => is this wanted?

source/ada/lsp-ada_documents.adb
```diff
@@ -2854,13 +2855,23 @@ package body LSP.Ada_Documents is
...
+            Provider.Propose_Completion
+              (Sloc   => Sloc,
+               Token  => Token,
+               Node   => Node,
+               Filter => Filter,
+               Names  => Names,
+               Result => Result);
+
+         exception
+            when E : others =>
```
AdrienBoulanger, line 2868: should we purge Result/Names or save them before and restore them in case of error?

testsuite/ada_lsp/completion.use_clauses/test.json
```diff
@@ -0,0 +1,273 @@
...
+               "id": 6,
+               "result": {
+                  "isIncomplete": false,
+                  "items": [
+                     {
+                        "label": "use Ada.Text_IO;",
+                        "kind": 11,
+                        "documentation": "Insert the use-clause corresponding to the with-clause on the same line.",
+                        "preselect": false,
+                        "sortText": "+use Ada.Text_IO;",
```
AdrienBoulanger, line 221: Does prefixing by "+" is enough for it to be shown first?

AdrienBoulanger (changes requested):
Some small questions/fixes.
Another test for "with A, B, C;" would be nice.

source/ada/lsp-ada_documents.adb
AnthonyLeonardoGracio: Yes, because it can happen in the case of use-clause completion. So I want to call the completion providers in this case, and let them handle the "Parent = null" case themselves (I have added a proper exception handler also, to prevent crashing the whole completion in case one provider fails).

source/ada/lsp-ada_documents.adb
AnthonyLeonardoGracio: No, because the exception handler is here to allow having results even in an exception happened (e,g: when a minor completion provider had an exception we want to have the results of all the providers that have succeed).

source/ada/lsp-ada_completions-use_clauses.ads
AnthonyLeonardoGracio: Done

source/ada/lsp-ada_completions-use_clauses.adb
AnthonyLeonardoGracio: Done

source/ada/lsp-ada_documents.adb
AdrienBoulanger: fine, the exception handler is a bit to wide: could we reduce it to Property_Error?

PR #1132, opened 2023-03-09
Add ada.selectProject command to select project

TamaMcGlinn:
This adds a filedialog command called 'Ada Select GPR Project' which lets the user select a .gpr file and puts that into the workspace configuration.

For a test project, see https://github.com/TamaMcGlinn/test_multigpr

Tested this with latest Ada_language_server (23.0.14) and it works beatifully; even if you had no .vscode/settings.json, it is created if you run the command. Note this command is only needed for non-alire projects, as ALS autodetects the right gpr file for alire projects now.


source/ada/lsp-ada_handlers.adb
```diff
@@ -792,8 +792,8 @@ package body LSP.Ada_Handlers is
          Self.Show_Message
            (VSS.Strings.To_Virtual_String
               ("More than one .gpr found." & Line_Feed &
-                 "Note: you can configure a project " &
-                 " through the ada.projectFile setting."));
+               "Configure using the ada.chooseProject command" & Line_Feed &
+               " or the ada.projectFile setting."));
```
TamaMcGlinn, line 796: On second thought, we probably don't want to mention vscode specific stuff here. `ada.projectFile` is actually what you need to set in all LSP clients.

TamaMcGlinn:
Added this note in the readme under VSCode > project support

> If there is a single .gpr file in the root of the project, it will be selected automatically. To choose the active project file, use the `ada.chooseProject` command and select a project file in the file dialog.

reznikmm:
@TamaMcGlinn 
I've tried it locally, but when I execute command the setting doesn't save in `.vscode/settings.json` immediately. So ALS doesn't get didConfigurationChange and "Build current project" task doesn't get project name.

Any idea?

TamaMcGlinn:
I had not tested that; good find! I will have more time to test this next week, I'll get back to you when I know if I can reproduce and/or fix it.

My own test was using test_multigpr, and just checking from the main.adb where cntrl+click on the Foo function call takes you. With that I could see that the language server was switching as soon as the command had been issued.

PR #1230, opened 2024-12-20
Fix invalid regex

### LINKED ISSUE #1229: [Bug]: Invalid regex in Ada TextMate grammar
### Environment

- OS and Version: N/A
- IDE Version: N/A
- Ada & SPARK Extension Version: Latest

### Bug Summary and Reproducer

Bug Summary:

[This line](https://github.com/AdaCore/ada_language_server/blob/1b326aac6a7bec689d9ef84e8236d4c573786663/integration/vscode/ada/advanced/ada.tmLanguage.json#L1705) is an invalid regex pattern (`"(?=(;|\\))"`), since it contains an unclosed group. It should include a third `)` at the end, since the second-last closing paren is escaped and matches a literal `)`.

(Note that the correctly-closed version of this pattern (`"(?=(;|\\)))"`) appears four other times in the same file. It is only this one instance that is missing the final closing paren.)

The result of this bug is that the pattern is an invalid Oniguruma regex, and it fails silently in `vscode-textmate`. It also causes the Ada grammar to error when using Shiki's [JavaScript engine](https://shiki.style/guide/regex-engines), since by default its JS engine doesn't silence regex errors.

### Configuration and Logs

```shell
N/A
```

### Other VS Code Extensions

_No response_

### Additional context

_No response_

slevithan:
Fixes #1229. This regex is invalid because it contains an unclosed group. As a result, it throws an error in Oniguruma (`vscode-textmate` silences regex errors). The corrected version of this regex is already used several times elsewhere in the grammar.

### DIFF
--- a/integration/vscode/ada/advanced/ada.tmLanguage.json
+++ b/integration/vscode/ada/advanced/ada.tmLanguage.json
@@ -1702,7 +1702,7 @@
             "patterns": [
                 {
                     "begin": "(?i)\\bwith\\b",
-                    "end": "(?=(;|\\))",
+                    "end": "(?=(;|\\)))",
                     "beginCaptures": {
                         "0": { "name": "keyword.ada" }
                     },


reznikmm:
I edited commit message, so GitHub doesn't understand the [changed commit](https://github.com/AdaCore/ada_language_server/commit/9329f709cf1aef897bfca679c54e0c3d17f2205f).

PR #1240, opened 2025-03-05
fix invalid command in build instructions

fatso83:
I tried the supplied command, it failed, I checked the help docs and found config, tried that and I could see that my config file now was changed. so it seems the old instructions are wrong.

### DIFF
--- a/doc/devel/build.md
+++ b/doc/devel/build.md
@@ -7,7 +7,7 @@
 
 1. Disable dependency sharing in Alire
 
-       alr settings --global --set dependencies.shared false
+       alr config --global --set dependencies.shared false
 
 1. Clone repository
 


eliericha:
Hello @fatso83 ,

Thank you for the contribution. However `alr config` is a deprecated Alire command and `alr settings` is the replacement. So the current build instructions are up to date.

This change occurred in Alire 2.0 (search for `alr config` in the [Alire changelog](https://alire.ada.dev/docs/)) which is the documented version for building. Perhaps you were using an older Alire version?

fatso83:
indeed. sorry for that. got `alire` from the current ubuntu repos. now getting from the relases on github.

Issue #1241, opened 2025-03-05
State memory requirements for building (much more than 8GB) or ways to circumvent OOM


fatso83:
The build script stopped with the following error the first time I ran it:
```
2 -O1 -gnatec=/tmp/GNAT-TEMP-000153.TMP -gnatec=/home/oscko/dev/ada_language_server/subprojects/libadalang/build/gnat.adc -gnatem=/tmp/GNAT-TEMP-000158.TMP /home/oscko/dev/ada_language_server/subprojects/libadalang/extensions/src/libadalang-unit_files.adb
gcc: fatal error: Killed signal terminated program gnat1
compilation terminated.
gcc: fatal error: Killed signal terminated program gnat1
compilation terminated.
gcc: fatal error: Killed signal terminated program gnat1
compilation terminated.
gcc: fatal error: Killed signal terminated program gnat1
compilation terminated.

   compilation of libadalang-data_decomposition.adb failed
   compilation of libadalang-doc_utils.adb failed
   compilation of libadalang-expr_eval.adb failed
   compilation of libadalang-preprocessing.adb failed

gprbuild: *** compilation phase failed
error: Command ["gprbuild", "-s", "-j0", "-p", "-P", "/home/oscko/dev/ada_language_server/subprojects/libadalang_tools/src/lal_tools.gpr", "-XVERSION=", "-XBUILD_DATE=", "-m", "-j0", "-m", "-vh"] exited with code 4
error: Compilation failed.
make: *** [Makefile:108: all] Error 1
error: Command ["make", "VERSION=", "GPRBUILD_CARGS=-m -vh", "all"] exited with code 2

ada_language_server on edge [?] took 7m18s
❯

```

The important part is the `Killed signal terminated program `. This typically happens when you are out of memory, I found out. I had 8GB of RAM and 2GB of swap, which is usually plenty. There is also nothing stating I need that much, but I added some additional swap space and it faired a bit better before being killed again:

![Image](https://github.com/user-attachments/assets/bc33ea3e-92ad-488e-8d87-04c1ced64b90)
![Image](https://github.com/user-attachments/assets/71a828c8-cde6-422c-865b-46776d86b09e)
![Image](https://github.com/user-attachments/assets/5334750a-e91a-4e1d-9970-2bba61df0725)

[...]

PR #1242, opened 2025-03-06
Add instructions to use official release builds

fatso83:
The current installation instructions make no mention of the release builds, telling new users to build the server from scratch, which seems a bit unnecessary ... Ref #1241


eliericha:
Thank you for the contribution @fatso83 !

PR #1301, opened 2026-06-29
Integer identifier

Vinuthnasupriyayadavalli:
To be more precise, the Normalization Form KC of the Unicode standard is applied to identifiers.


godunko:
spam/AI

